Issue #31, opened 2023-02-22
Setting as Primary or Secondary is not persisting


Teddarific:

zachsnoek:
Hey, @Teddarific –– I'd love to take this. [Is this](https://www.loom.com/share/6a0c392eae12442d865b87058c33c030) the bug you're describing? If not, could you provide repro steps?

Teddarific:
Hey @zachsnoek! That is exactly the bug - would love if you took a crack at it, feel free to open up a PR and tag me in it. Thanks for contributing 😍

Teddarific:
Fixed by @zachsnoek in #51

PR #51, opened 2023-02-27
fix: ensure primary/secondary color persists

zachsnoek:
# Summary

This PR fixes the primary/secondary color persistence issue reported in #31.

Current color editor behavior: https://www.loom.com/share/6a0c392eae12442d865b87058c33c030
Behavior with fixes in this PR: https://www.loom.com/share/50b6590fce7f49d68f859e224b5cabd6

# Description

`<EditColorModal />` has two stateful `isPrimary` and `isSecondary` values whose [initial values come from `initialColorData`](https://github.com/zachsnoek/mirrorful/blob/3c7fa2ede409c380d33031f5136d9a4dc5e405d5/packages/mirrorful/editor/src/components/ColorPalette/EditColorModal.tsx#L54-L59). When the modal closes, [they get passed to `onClose`](https://github.com/zachsnoek/mirrorful/blob/3c7fa2ede409c380d33031f5136d9a4dc5e405d5/packages/mirrorful/editor/src/components/ColorPalette/EditColorModal.tsx#L136-L137), which updates the color data.

When `color.isPrimary` and `color.isSecondary` get updated via the menu, the corresponding `<EditColorModal />` state isn't updated, so `onClose` gets called with their initial values, and we get the behavior shown above.

Since there isn't really a need for local `isPrimary` and `isSecondary` states in the modal currently, we can fix this by deriving the values from `initialColorData` instead.

I also removed some dead code and imports.

### DIFF
--- a/packages/mirrorful/editor/src/components/ColorPalette/ColorPaletteSection.tsx
+++ b/packages/mirrorful/editor/src/components/ColorPalette/ColorPaletteSection.tsx
@@ -44,7 +44,7 @@ export function ColorPaletteSection({
 
                 newColors.forEach((color) => (color.isPrimary = false))
                 newColors[colorIndex].isPrimary = true
-                console.log(newColors)
+
                 onUpdateColors(newColors)
               }}
               onSetAsSecondary={() => {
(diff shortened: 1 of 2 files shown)


Teddarific:
Great catch Zach — thanks for cleaning up the code as well! Merging :)

PR #62, opened 2023-03-01
i18n - german translation of main README.de.md

tobiasdossinger:
- adds German top-level README.de.md
- links from top-level README to German README.de.md

**Notice**
Missing the following two READMEs --> `package/mirrorful` directory, and the UI itself in `package/mirrorful/editor`. 
I would suggest putting it in a follow-up ticket when the i18n gets setup. 

I am curious about your feedback. What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,6 @@ No matter the PR, all the Github avatars of contributors will be added to the Mi
 
 ## 🌎 Translations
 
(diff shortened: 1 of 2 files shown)


alexdanilowicz:
Fantastic! This is great. Danke! 

Going merge this!

Issue #57, opened 2023-02-27
Recommend `hover` & `active` variants of colors


Teddarific:
![image](https://user-images.githubusercontent.com/16343600/221658611-2319bac1-1841-48a5-a252-beb802afc273.png)

Adding a new color right now can be intimidating / confusing because of the `hover` & `active` options. We should add a "magic wand" icon button that will prefill in a recommended shade. We can just prefill in `colorShades['600']` as the `hover`, and `colorShades['700']` as the `active`

alexdanilowicz:
Hey @zachsnoek! Thanks so much and nice to meet you! Go for it. You're on fire 🔥 

Let us know if we can be helpful.

PR #65, opened 2023-03-01
feat: add recommended shade buttons

zachsnoek:
Demo: https://www.loom.com/share/d721fd16dcdd49fc8123c5ac4fa45bc9

This PR addresses #57: It adds magic wand icon buttons in the hover and active inputs, allowing users to fill the input with a recommended shade.

I also noticed that the `<ColorPicker />` alignment is a bit funky because the underlying `<SketchPicker />` has `box-sizing: content-box`. I changed it to `border-box` so that it's easier to work with and updated the modal layout to two equally-sized columns.


Teddarific (approved):
Looking great! Thanks for taking the time to form `<InputGroup />`s! 

I'll take a pass at a QA later today, and then merge it in! Killing it @zachsnoek!

PR #75, opened 2023-03-02
Add new color skeleton

prichter:


prichter:
**Summary**
Update based in issue-24.  3 main edits:
1. Create component AddColorRow
2. Update utils to re-use handleInvalidColor from EditColorModal
3. Add newly created component in ColorPaletteSection

**Description**
Request to add skeleton instead of Add New Color button in editor.  

This required changes from the ColorRow component that had been previously used as demo (see issue comments) so the options were to update the ColorRow component to make it configurable for blank row, copy-paste code from this component into ColorPaletteSection, or create new component.  The latter was selected to reduce chance of breaking existing components and allow for further customization if needed.

Wasn't sure what color styling was going to be needed, so component requires passing in of the background color (via 'backgroundColorString').  To prevent invalid css coming in, code was pinched from EditModalDialog to validate.  Rather than duplicating in component, I updated utils.ts with this code for export (this could be reused in a later EditModalDialog update).

New component placed into ColorPaletteSection is contained in BOX component which handles the onClick event calling the same function as the former "Add New Color" button to open the EditColorModal dialog.  Styling of new component was based on feedback from @Teddarific  and @alexdanilowicz.

Happy to take feedback if this is too complicated.

alexdanilowicz:
I think @Teddarific refactored the code a bit in 3.0.0, so will let him merge this/review this/handle this.

prichter:
OK I synced with main and pushed changes to make text not-centered.  I got confused with the syncing - I think it's OK, but if it needs changes, please let me know.

Teddarific:
Hey Paul! Thanks for re-syncing with `main` — let me take a look. The screenshot looks great!

prichter:
Closing pull request, will refactor based on update.

Issue #59, opened 2023-02-28
If a color is already set as primary or secondary, disabled the button and change it to say "Already set as primary"


alexdanilowicz:
<img width="1340" alt="Screen Shot 2023-02-27 at 8 52 47 PM" src="https://user-images.githubusercontent.com/29822597/221757688-4db9dc28-4910-404d-94f3-4a4f8ddbe2a8.png">

lagneshthakur:
I can help in fixing this.

aryanraj60:
Hello! I would like to take ownership of this issue and work on resolving it. Could you please assign it to me? Thank you!

alexdanilowicz:
@lagneshthakur to get started, after forking the repo, you can run the following commands:

```
➜  mirrorful git:(main) ✗ cd packages/mirrorful/editor
➜  mirrorful git:(main) ✗ git checkout -b "lagneshthakur/issue-59"
➜  mirrorful git:(paul/issue-53) ✗ yarn; cd editor; yarn; cd .. # install packages
➜  mirrorful git:(paul/issue-53) ✗ yarn run local # runs mirrorful locally

$ rm -rf ./dist; tsc --build; cross-env NODE_ENV=development node ./dist/cli/index.js
Running Mirrorful version: 2.0.19
Starting Mirrorful in /Users/zeus/code/mirrorful/packages/mirrorful.


Visit: http://localhost:5050/
Inside your project, you can run:

  yarn run mirrorful

to start the visual editor at any time 🚀
```
From there, you can then make changes in code and see the changes hotreload.

lagneshthakur:
@alexdanilowicz Thank you, this is exactly what I was looking for.

alexdanilowicz:
@lagneshthakur let us know if you need help!

PR #77, opened 2023-03-02
resolves https://github.com/Mirrorful/mirrorful/issues/59

lagneshthakur:
Summary
Pull request to fix issue 59. This PR adds check for isPrimary and isSecondary on the buttons and changes text based on this condition.

Description
When a color is Set as Primary, it's attribute isPrimary is set to true, and based on this attribute, we disable the 'Set as Primary' button and also change it's text to 'Already Primary'. Same happens for Set as Secondary button.

Here's a small video showing how it looks on screen - 
https://watch.screencastify.com/v/tHwXBcOyT7rnnqppjffq


alexdanilowicz:
Great work @lagneshthakur!

Issue #79, opened 2023-03-02
Add an empty `Add color variant` placeholder square


Teddarific:
![image](https://user-images.githubusercontent.com/16343600/222566942-3ee7c948-5c36-4ed7-bcef-454a7bd5301a.png)

prichter:
Happy to do this if tomorrow is OK, please assign?

prichter:
Working on this...

alexdanilowicz:
Assigned! Go 🇦🇺 !!! haha

PR #81, opened 2023-03-02
Issue #24 - Add new color skeleton

prichter:
**Summary**
This PR addresses issue #24 by adding an 'AddColorSkeleton' component to the project, and adds this into the `ColorPaletteSection` component.

**Description**
New component `AddColorSkeleton` added, which borrows from the new layout of the color row.  It has two variables, `backgroundColorString`, allowing you to set the background colour of the mockup (displayed with low opacity), and `numberOfMockVariants` being the number of mock variant squares to be displayed by the skeleton.

The whole skeleton is added within a box UI component in the `ColorPaletteSection` which, when clicked, opens the new color modal dialog, as per the previous button functionality.

There is a `handleAddColorCss` function that adjusts the skeleton's spacing depending on whether other colors are displayed or not.

Feedback welcome.

PR #84, opened 2023-03-03
Add empty placeholder square for new variant

prichter:
** Summary **
Created `ColorVariantPlaceholder` component for new variant placeholder as requested in issue #79, and added to `ColorRow` component.

**Description**
Created component `CollorVariantPlaceholder` which emits onClick event and accepts backgroundColorString to set the control background color.  This uses available Chakra-ui/icons AddIcon for the + icon, as this was easiest to achieve in short order, as shown below.

This component was added to the ColorRow component with the click event calling the `onAddVariantModalOpen()` method.

image


alexdanilowicz:
@prichter if you have time, can you make the background color look more like what the "Edit Color Name" and "Add Variant" look like. It would be antithetical to Mirrorful's mission if we used a raw string like `#eee` as our color haha. Instead of passing a backgroundColor prop, we should probably just use a Charka button but make it wide as you have it.

If a Charka button is not flexible enough, then I think you can set the Box colorScheme to like `gray` or something like that.

alexdanilowicz:
Updated this branch with main, so you want to pull before you start working on it again!

alexdanilowicz:
Made some changes because at first I thought it looked like the light grey is another color. I think this is a good start, but honestly I'm not sold on it yet. It feels like a lot of empty space just for a button. Going to push up my changes, I think this is a really good start, but going to hold off merging.

image

@Teddarific FYI if you want to take this across the finish lines or have thoughts. Down to wait.

alexdanilowicz:
Oops wrong PR, deleting my above comment. I meant to post on #81

alexdanilowicz:
image

Cleaned it up. @prichter in case you are interested in the changes: https://github.com/Mirrorful/mirrorful/pull/84/commits/a3ef870c3d938d262a5e36cf835834730a5862d8

Great work!

Issue #80, opened 2023-03-02
When adding colors with same name, auto increment stops at 2


prichter:
If you add several colors with the same name, initially the first duplicate is named `[color name] 2`, but the second duplicate is also named `[color name] 2`, which results in both disappearing if one is deleted.

prichter:
Happy to do this tomorrow, please assign

alexdanilowicz:
Go for it! Thanks @prichter

PR #85, opened 2023-03-03
Issue #80 - Auto increment color suffix number on color duplication

prichter:
**Summary**
Implement code in ColorPaletteSection to increment the suffix on an added color if it has been used before.  This simply maps the existing colors, performs a regex match, and analyses if matched to determine the max. suffix value.

**Description**
`ColorPaletteSection` component EditColorModal onClose method was updated to check passed in color name against each existing color.  Existing colors were iterated (using .map) and each color was tested using a regex match.  The third part of the regex match is effectively the suffix if there a color has been added multiple times with the same name.  This was tracked to determine if it was the largest increment, and if so, this is added to the color name as a suffix.


alexdanilowicz:
This makes sense to me, will QA tomorrow but thanks for opening the PR! @prichter

alexdanilowicz:
QA-ed! Merging!

Issue #82, opened 2023-03-03
When adding new variant, invalid color value is not handled


prichter:
If you add a new color variant to a color, and add gibberish or an invalid CSS color, there is no warning, and variant appears as white-on-white.  Suggest validation error (eg input turning red) on invalid css submission.

PR #86, opened 2023-03-03
Issue #82 - Handle invalid colors in variant modals

prichter:
**Summary**
This PR is a fix for issue #82.  Previously, saving a new or edited variant color would permit, and save, invalid CSS color values.  This fix adds error checking and Input highlighting prior to save, following the same convention as in the onboarding name page.

**Description**
Implementation is the same in both the `AddVariantModal` and the `EditVariantModal` in that an error variable is created, and used to highlight incorrect / missing syntax on the color inputs within the modal.  A new Text Chakra component is added beneath each input to display relevant error message.

The validation is done by saving the initial value, and applying the handleInvalidCss util to the input color, then checking for differences, and setting the error message in the event that there is a difference (ie. the handleInvalidCss has changed the value).

### DIFF
--- a/packages/mirrorful/editor/src/components/ColorPalette/AddVariantModal.tsx
+++ b/packages/mirrorful/editor/src/components/ColorPalette/AddVariantModal.tsx
@@ -12,7 +12,7 @@ import {
   FormLabel,
   Box,
   Checkbox,
-  Text
+  Text,
 } from '@chakra-ui/react'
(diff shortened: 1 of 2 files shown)


alexdanilowicz:
Just qa-ed, and pushed up some formatting because I don't think you have prettier set up. (DM-ed you about this). Fantastic work yet again.

PR #98, opened 2023-03-04
#98 - Adds nuxt 3 example

tobiasdossinger:
Just added the nuxt 3 example with a bit of eslint+prettier


tobiasdossinger:
#98

alexdanilowicz:
@tobiasdossinger checking this out now. Could you please add it in our README as an example under the examples section?

alexdanilowicz:
@tobiasdossinger. Also, I added a `.mirrorful` folder that is exported. Could you add that actually being imported and used in the nuxt 3 app? because I think Nuxt + mirrorful simply added as a dev dependency is not enough of an example. We should show mirrorful actually being consumed and used in a component. Thanks!

tobiasdossinger:
@alexdanilowicz 

**Just added an example component with two variations of using Mirrorful within Nuxt 3:**
1. Using the generated color vars within tailwind config
2. Using the vars inside the component as they are generated

Disclaimer: The Token Variation is still missing in this example.
 
Let me know what you think.

alexdanilowicz:
Screen Shot 2023-03-05 at 3 46 04 PM

Looks great! Feel free to merge! 

(I'm curious do you have merge access after I approve the review, or do I need to merge?)

alexdanilowicz (approved):

alexdanilowicz:
Thanks so much for making all the changes @tobiasdossinger!

tobiasdossinger:
> Screen Shot 2023-03-05 at 3 46 04 PM
> 
> Looks great! Feel free to merge!
> 
> (I'm curious do you have merge access after I approve the review, or do I need to merge?)

I don't know if I have merge access after the approval. Would be nice to know.

Issue #105, opened 2023-03-07
Internationalization of Mirrorful Editor Dashboard with i18n


tobiasdossinger:
Adding i18n to the Mirrorful Editor

alexdanilowicz:
We should use https://nextjs.org/docs/advanced-features/i18n-routing NextJS i8n routing for this.

PR #118, opened 2023-03-09
#105 | Adds i18n to editor

tobiasdossinger:
#105

### DIFF
--- a/packages/mirrorful/editor/next.config.js
+++ b/packages/mirrorful/editor/next.config.js
@@ -5,6 +5,14 @@ const nextConfig = {
     // your project has ESLint errors.
     ignoreDuringBuilds: true,
   },
+  i18n: {
+    // These are all the locales you want to support in
+    // your application
+    locales: ["en", "de"],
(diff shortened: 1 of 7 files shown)


alexdanilowicz:
@tobiasdossinger JFYI, the dashboard has changed _a lot_ over the last few days and we're moving to a monorepo set up here: https://github.com/Mirrorful/mirrorful/pull/190

So, it might be really hard to keep i8n up while the rest of the product is moving quickly. I'll keep this PR up, but just wanted to set expectations, as I think right now we are focusing on landing more features first.

tobiasdossinger:
Thanks for the notice. I had very busy last weeks, so I just saw this today. Looking forward to getting back onto some issues and PR.

alexdanilowicz:
Yeah are you familiar with i8n and next? if you could set that up for the editor would be awesome! @tobiasdossinger

tobiasdossinger:
Not with Next, but with Nuxt, so it should be similar. I will set it up.

Issue #58, opened 2023-02-28
bug: Color labeled as unknown


alexdanilowicz:
<img width="811" alt="Screen Shot 2023-02-27 at 8 50 17 PM" src="https://user-images.githubusercontent.com/29822597/221757135-6065d168-21a5-4d53-8971-c5d6a693d34e.png">

aryanraj60:
Hello! I would like to take ownership of this issue and work on resolving it. Could you please assign it to me? Thank you!

alexdanilowicz:
Hey @aryanraj60 go for it! Please first join our slack channel: https://join.slack.com/t/mirrorful/shared_invite/zt-1ps2xtxh0-2NaixFfFzSKZbr5gw_AHfA

And after forking the repo, here's how you can get set up:

```
➜  mirrorful git:(main) ✗ cd packages/mirrorful/editor
➜  mirrorful git:(main) ✗ git checkout -b "aryanraj/issue-58"
➜  mirrorful git:(paul/issue-53) ✗ yarn; cd editor; yarn; cd .. # install packages
➜  mirrorful git:(paul/issue-53) ✗ yarn run local # runs mirrorful locally

$ rm -rf ./dist; tsc --build; cross-env NODE_ENV=development node ./dist/cli/index.js
Running Mirrorful version: 2.0.19
Starting Mirrorful in /Users/zeus/code/mirrorful/packages/mirrorful.


Visit: http://localhost:5050/
Inside your project, you can run:

  yarn run mirrorful

to start the visual editor at any time 🚀
```

Then, you can make changes in code, and once done, open a PR for your changes. Let us know if any questions!

alexdanilowicz:
Hi @aryanraj60 are you working on this? Let us know!

iamoghenerukevwe:
Can I take this up?

alexdanilowicz:
Sure go for it @iamoghenerukevwe !

iamoghenerukevwe:
Does it need to be assigned to me?

alexdanilowicz:
Going to clear the assignments, any one is welcome to open a PR for this!

alexdanilowicz:
@iamoghenerukevwe go for it!

ktra99:
I'm currently looking into this as well and the issue is that the `nameThatColor` function is not able to generate a name for a specific combination of `h, l, s`. Would it be possible to get more clarity on what the current combinations are based on? -

```
if (s <= 0.1 && l >= 0.9) {
[...]

PR #123, opened 2023-03-11
fix: Color labeled as unknown #58

iamoghenerukevwe:
Color labelled as unknown #58


alexdanilowicz:
Hey thanks Michael! @iamoghenerukevwe 

Good debugging here, could you explain why the hsl changes fix it?

I think it's better, but I still get an unknown if you regenerate a couple of times: 
image

This is for hex #ea7a96. I need to take a closer look. But perhaps you'll know the bug right away?

prichter:
OK, did some testing and I think the problem here is that tinycolor.toHSL() returns values as hue, saturation, lightness _in that order_, but the `nameThatColor` method takes parameters in _hue, lightness, saturation_ order.  I suspect that this is causing some out of range values to be assessed, resulting in "Unknown" color.

prichter:
OK, I have identified some gaps in the `nameThatColor` functionality.  The edge case is where `tinycolor().toHsl()` returns decimal values for HSL that are between the boundary conditions in the 'if' cases.  For example, if hue of 54.54 is returned, if falls between the <=54 and >=55 cases (brown/orange vs brown/yellow cases in `nameThatColor`).
The crux of the matter is that tinycolor returns decimal values, but `nameThatColor` expects integer values.

alexdanilowicz:
nice @prichter — thanks for chiming in. Really well explained. Excellent. Do you know the ultimate fix? Feel free to push it up

alexdanilowicz:
Great teamwork here.

iamoghenerukevwe:
Thanks @prichter for so coherently putting it, I discovered that boundary conditions in the if statements were causing the error of "Unknown" and to resolve the issue  would simply be to make sure that there are no gaps, in the values used for the if statements. I'll just hold until @prichter is done, then I can add in more color names.

alexdanilowicz:
image

This also seems messed up (this is on `main`). I'd expect Yellow vs Green for `#f1f30d`. 

Screen Shot 2023-03-14 at 3 21 33 PM

Same with red vs pink for `#f30d55`

I wish we could use `tinycolor.toName()` from https://github.com/bgrins/TinyColor but it seems that returns false for hex values. I bring up the above bugs in case the algo as a whole needs to be fixed, instead of just patching it

prichter:
@alexdanilowicz certainly.  I'll work on this next...

alexdanilowicz:
Closing this in favor of https://github.com/Mirrorful/mirrorful/pull/182

Nice work, @prichter !

Issue #127, opened 2023-03-11
After running npx mirrorful, immediately pop up localhost:5050 like they do with create react app


alexdanilowicz:
CRA - create react app - pops up localhost:5050 in your browser for you after running it. we should do the same. 

CRA is open-source. Need to see how they do it.

Ujjwal55:
I would like to take up this issue @alexdanilowicz

alexdanilowicz:
Hey @Ujjwal55, actually this has already been fixed in https://github.com/Mirrorful/mirrorful/pull/129. It seems Github did not automatically close the issue once the PR was merged. There are some other great first issues here if you're interested: https://github.com/Mirrorful/mirrorful/contribute

Welcome to the community 👋🏼 . Have you already joined our Slack? https://join.slack.com/t/mirrorful/shared_invite/zt-1ps2xtxh0-2NaixFfFzSKZbr5gw_AHfA

PR #129, opened 2023-03-12
Open localhost:5050 automatically when running mirrorful cli

karanvirsb:
Made a simple open browser feature following the create-react-app library. #127

### DIFF
--- a/packages/mirrorful/cli/helpers/openBrowser.ts
+++ b/packages/mirrorful/cli/helpers/openBrowser.ts
@@ -1,0 +1,12 @@
+import open from 'open'
+type props = { url: string; port: string }
+export default async function openBrowser({ url, port }: props) {
+  try {
+    console.log(`Opening up browser to ${url}:${port}`)
+    const res = await open(`${url}:${port}`)
+
+    console.log('🚀 ~ file: openBrowser.ts:8 ~ openBrowser ~ res:', res)
+  } catch (error) {
+    console.log('🚀 ~ file: openBrowser.ts:10 ~ openBrowser ~ error:', error)
+  }
+}
--- a/packages/mirrorful/cli/helpers/openBrowser.ts
+++ b/packages/mirrorful/cli/helpers/openBrowser.ts
@@ -5,7 +5,7 @@ export default async function openBrowser({ url, port }: props) {
     console.log(`Opening up browser to ${url}:${port}`)
     const res = await open(`${url}:${port}`)
 
-    console.log('🚀 ~ file: openBrowser.ts:8 ~ openBrowser ~ res:', res)
+    console.log('🚀 Opening Mirrorful in your browser! ', res)
   } catch (error) {
     console.log('🚀 ~ file: openBrowser.ts:10 ~ openBrowser ~ error:', error)
   }
--- a/packages/mirrorful/cli/helpers/openBrowser.ts
+++ b/packages/mirrorful/cli/helpers/openBrowser.ts
@@ -22,7 +22,5 @@ async function waitToOpenBrowser(port: number) {
     )}`
   )
   console.log()
-  let timer = setInterval(() => {}, 1000)
   await sleep(5000)
-  clearInterval(timer)
 }
--- a/packages/mirrorful/cli/helpers/openBrowser.ts
+++ b/packages/mirrorful/cli/helpers/openBrowser.ts
@@ -4,29 +4,25 @@ import sleep from './sleep'
 type props = { url: string; port: number }
 export default async function openBrowser({ url, port }: props) {
   try {
-    await waitToOpenBrowser()
+    await waitToOpenBrowser(port)
     await open(`${url}:${port}`)
   } catch (error) {
     console.log(
-      'Go to localhost:5050 to open Mirrorful. Could not open automatically: ',
-      error
+      `Visit: ${chalk.green(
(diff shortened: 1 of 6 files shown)


alexdanilowicz:
Hey @karanvirsb! Thanks for jumping in, but this isn't what I intended from the issue. I was thinking of adding a line in the mirrorful cli script (code found here: https://github.com/Mirrorful/mirrorful/tree/main/packages/mirrorful/cli) or in the package json, as you have it, that prompts the opening of localhost:5050 even if it's not already running.

in other words:

```
npx mirrorful # should open up localhost:5050 in the browser
```

I believe you code right now will only work if localhost:5050 is already running?

karanvirsb:
Hey, @alexdanilowicz I followed the create-react-app react-scripts here: https://github.com/facebook/create-react-app/blob/d960b9e38c062584ff6cfb1a70e1512509a966e7/packages/react-dev-utils/openBrowser.js#L140. I found that they used the open library to help them open the browser. If I move the node ./openBrowser.mjs from the script and add it to a different line in the package.json like this:  

```javascript
"scripts": {
    ...
    "openBrowser": "node ./openBrowser.mjs",
    ...
}
```

It will still open the browser and show this: 

image

karanvirsb:
Hey, @alexdanilowicz I have moved openBrowser.ts to the helpers in the cli and added it to the Mirrorful cli init script and it seems to work well. If we moved around the openBrowser function within the init script it determines how early the browser will open.

packages/mirrorful/cli/init.ts
```diff
@@ -31,7 +32,7 @@ export async function init({
   await makeDir('.mirrorful')
 
   const port = 5050 // don't hard code this
-
+  await openBrowser({ url: 'http://localhost', port })
```
alexdanilowicz, line 35: Awesome! I think this line should be moved to the end. I think it should be on line 107 instead, after ensuring there are no errors

packages/mirrorful/cli/helpers/openBrowser.ts
```diff
@@ -0,0 +1,12 @@
+import open from 'open'
+type props = { url: string; port: number }
+export default async function openBrowser({ url, port }: props) {
+  try {
+    console.log(`Opening up browser to ${url}:${port}`)
+    const res = await open(`${url}:${port}`)
+
+    console.log('🚀 ~ file: openBrowser.ts:8 ~ openBrowser ~ res:', res)
```
alexdanilowicz, line 8:
```suggestion
    console.log('🚀 Opening Mirrorful in your browser! ', res)
```

packages/mirrorful/cli/helpers/openBrowser.ts
```diff
@@ -0,0 +1,12 @@
...
+  } catch (error) {
+    console.log('🚀 ~ file: openBrowser.ts:10 ~ openBrowser ~ error:', error)
```
alexdanilowicz, line 10:
```suggestion
    console.log('Go to localhost:5050 to open Mirrorful. Could not open automatically: ', error)
```

alexdanilowicz (commented):
few nit comments

alexdanilowicz:
Hey @karanvirsb awesome! incredible, thanks so much for your time and help with this! Left a few comments.

packages/mirrorful/cli/helpers/openBrowser.ts
karanvirsb: I think maybe we should just remove the result variable from the log as it might bloat up the console.

alexdanilowicz:
By the way, @karanvirsb, are you in our [Slack community](https://join.slack.com/t/mirrorful/shared_invite/zt-1ps2xtxh0-2NaixFfFzSKZbr5gw_AHfA)? 

Gave you a shoutout here:
image

packages/mirrorful/cli/init.ts
karanvirsb: If we move it all the way to the bottom it seems to not open the browser.

alexdanilowicz: Oh interesting, makes sense

alexdanilowicz:
Hmmm, I think the lag of the site not being available thought is not a good user experience.

Demostration here:
https://www.loom.com/share/9734c09251f347deba19540086ca3e33

Really appreciate the hard work on this, but I think we either 

1) Need to ensure it doesn't even pop up when not available.. perhaps put it in some sort of timeout? That might be too hacky though. But I think it would be good to see the console output too.
2) If we can that to work, not a big deal as we are probably going to move to a cloud editor soon.

karanvirsb:
I think the delay is mostly due to the next dev and the spawn child sync. I tried fixing it by making the spawn async. setTimeout might be too hacky since you need to know how long the next dev takes to run.

alexdanilowicz:
Nice, I like it and think it's a fantastic improvement. I made a few changes to the language that is outputted. Let me know what you think. 

@Teddarific — do you have thoughts here? I would prefer to release `4.0.1` first, and then we merge this change to be released with `4.0.2`.

karanvirsb:
> Nice, I like it and think it's a fantastic improvement. I made a few changes to the language that is outputted. Let me know what you think.
> 
> @Teddarific — do you have thoughts here? I would prefer to release `4.0.1` first, and then we merge this change to be released with `4.0.2`.

Nice, I think it looks good, I like the changes!

PR #131, opened 2023-03-12
Fix test issue with random hex generation

prichter:
**Summary**
During initial testing it was discovered that, about 5% of the time, the function `handleInvalidColor` in `/ColorPalette/utils.js` is returning values which are shorter than 6 digits.  This fixes the problem by generating 6 hexadecimal characters and appending them to a string, consistent with regex checks.

**Detail**
The function handleInvalidColor generates a random hex number, but because of the implementation of the range of valid numbers, sometimes a four- or five-digit number is returned (presumably being at the smaller end of the random number range), which is inconsistent with the regex test used at the start of this function.  This fix, whilst less efficient due to the repeated loop of Math.floor, ensures the validity of the hex code by creating a string of 6 random hex numbers, rather than randomising to obtain a whole value.  Given this is largely only expected to be used during perhaps onboarding, and dealing with bad values being fed in, I believe any performance hit should be acceptable.


prichter:
Retested wth Jest, 500 repetitions, all OK:
image

alexdanilowicz:
Interesting! Thanks for opening @prichter. I actually got the original implementation from this blog post: https://www.paulirish.com/2009/random-hex-color-code-snippets/ and seemed legit to me. Do you know why the original fails? 

If your new solution passes tests, I'm all good with it. I'm just kind of surprised to see that the original implementation is no good, but I guess you can't trust everything you read on the Internet, hah.

Side note: we need to change the logic in that file in general because `rgb()` values should be valid too. It's really bad how a user can copy paste in an RGB value, but then it changes to a random hex

packages/mirrorful/editor/src/components/ColorPalette/utils.ts
```diff
@@ -71,7 +71,11 @@ export const handleInvalidColor = (input: string) => {
   }
 
   const validSubsetRegex = /^#?[0-9A-Fa-f]{0,6}$/ // regex to validate if input is a valid subset of a hexcode
-  const randomHex = Math.floor(Math.random() * 16777215).toString(16) // generate a random valid hexcode
+
+  let randomHex :String = ''
+  for(let i:number = 0; i < 6; i++){
```
alexdanilowicz, line 76: let's try and figure out why your `prettier` is still not working (be sure to install extension + turn on autoformat on save), but this is ✅

alexdanilowicz (approved):

alexdanilowicz:
Merging this!

PR #133, opened 2023-03-12
Add jest for unit testing

prichter:
Do not merge - draft only for discussion.


alexdanilowicz:
This looks pretty good, very nice. @prichter. Will hop on it later today because ideally we keep the PR small and get this merged in. Love this initiative, testing is so important! 🙌🏼

PR #135, opened 2023-03-12
change default base color if  is checked when adding a new variant

### LINKED ISSUE #130: On adding new variant, set as base checkbox not working
If you add a new variant, and select "Set as Base" checkbox prior to saving, the new variant is not saved as the base variant.  Instead, it is saved as a regular variant (see attached Loom for more details):
https://www.loom.com/share/b088275605364c16a5c4a76a3e697de4

ktra99:
resolves #130


alexdanilowicz (approved):
Awesome! That was quick, thanks so much!

PR #136, opened 2023-03-12
add regex support for rgb values

### LINKED ISSUE #134: Allow rgb values and hexes when adding a new color
As a user, I want to copy paste in RGB values. But right now, the add / edit color logic in the modal does crazy logic where if it's not a hex, it assume you're typing in a valid hex. We should keep that logic, but if it's RGB, allow RGB. 

For example in the valid color form:

rgb(255, 255, 255) should be a valid value

ktra99:
resolves #134


alexdanilowicz (approved):
Amazing! Great work!

alexdanilowicz:
@ktra99 — FYI I made two changes:

1. I sanitize the input first, as the regex would fail if there were spaces. `rgb(255,255,255)` and `rgb(255, 255, 255` (with spaces) should both pass

2. I removed support for alpha values, as I'm not sure that will work in export format. Need to do some testing.

PR #138, opened 2023-03-12
add to export description that a .mirrorful folder was created

### LINKED ISSUE #128: Export Success Modal / Import Instructions - Make more clear that a .mirrorful folder was created
In our Export Success Modal instructions and Import instructions we should make more clear that a .mirrorful folder is created

ktra99:
resolves #128


alexdanilowicz:
Nice! @ktra99 could you also update this in the `ImportInstructions.tsx` file as well. 

Also, let's change language to:

```
(actual path may vary. You can reference the <Code>.mirrorful</Code> folder in the root of your project)
```

Let me know what you think of that?

Thanks!

PR #139, opened 2023-03-12
implemented delete dialog for color, color variant and font size variant

### LINKED ISSUE #137: When deleting a color it should prompt the user to confirm
When a user presses delete a colour palette it is immediately deleted. This is the same issue with variants as they can be deleted immediately too. 

I think we should have a confirmation prompt to ask the user to make sure they want to delete the palette or even a variant.

ktra99:
resolves #137


alexdanilowicz:
Nice! In the future, a Loom/screenshot would be very helpful, but I hopped on the branch and made some changes. Fantastic work @ktra99. Looks good. Also FYI, just to give you a heads up in the future, we may not ultimately merge/do every issue opened by a community member (such as #137), so be on the lookout for maintainers (me or @Teddarific) assigning it or discussing it before doing the issue to ensure it is within Mirrorful's roadmap, but I think is a great addition. Thank you for doing it! 

Great, great, GREAT work! 🥳

Issue #142, opened 2023-03-13
Clear add color modal state on close


zachsnoek:
Demo: https://www.loom.com/share/1080d75392504e6ea7b73c5f3d0bc87b

Adding multiple colors is difficult because the modal state doesn't reset. See [this Slack message](https://mirrorful.slack.com/archives/C04QRA1GA4C/p1678668860067089?thread_ts=1678663135.944999&cid=C04QRA1GA4C).

alexdanilowicz:
This is resolved.

PR #143, opened 2023-03-13
fix: reset color modal state on close

zachsnoek:
This PR addresses #142: it clears the `<EditColorModal />` state on close.

Before: https://www.loom.com/share/1080d75392504e6ea7b73c5f3d0bc87b

After: https://www.loom.com/share/d6ecb2261f90403da520f6b390cb01d5

### DIFF
--- a/packages/mirrorful/editor/src/components/ColorPalette/EditColorModal.tsx
+++ b/packages/mirrorful/editor/src/components/ColorPalette/EditColorModal.tsx
@@ -12,17 +12,14 @@ import {
   FormLabel,
   Box,
   Flex,
-  InputRightElement,
-  IconButton,
-  InputGroup,
 } from '@chakra-ui/react'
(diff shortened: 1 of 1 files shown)


alexdanilowicz:
Nice! I think we could also use a useEffect that resets the state if the "open" field changes.

you implementation works too, well done!

zachsnoek:
> Nice! I think we could also use a useEffect that resets the state if the "open" field changes.

True! Though I'll note that using an effect for that is usually [unnecessary](https://beta.reactjs.org/learn/you-might-not-need-an-effect) and makes the code more complicated and harder to follow.

alexdanilowicz:
Brilliant! And this why I do open-source... to learn something new! Thank you for sharing and appreciate you linking that resource. Makes total sense. Merging this in. 

image

Issue #125, opened 2023-03-11
Clarify that you don't need to add a hyphen in variable name 


alexdanilowicz:
As a user, when I'm naming a color, I'm confused it if must name it `primary-blue` because I'm used to variables not having spaces.

Solution: add a tooltip or make it more clear in some other way

arneesh:
Hey @alexdanilowicz, I have given a merge request for this issue: [#144 ](https://github.com/Mirrorful/mirrorful/pull/144)

PR #144, opened 2023-03-13
Added tooltip for variable naming

arneesh:
Resolves: [#125 ](https://github.com/Mirrorful/mirrorful/issues/125)

- Added a tooltip for variable naming stating that hyphens are not necessarily required in variable naming.


alexdanilowicz:
Awesome @arneesh, in the future, are you able to attach a screenshot or video (or Loom) to the PR? Thank you! Otherwise, looks good I'll get this merged in soon.

alexdanilowicz:
Some feedback:

I think instead of this
image

We should wrap the tooltip in an icon, like so:

```javascript
import { InfoIcon } from '@chakra-ui/icons'

<Tooltip ...>
  <InfoIcon />
</Tooltip>
```

That way the user only reads it IF they are looking for more info.

Let's also change the props/language to:

```
              <Tooltip placement="right" closeDelay={500} hasArrow label={"Variable names don't need a hyphen."}>
```

Thoughts @arneesh ?

arneesh:
Agreed, @alexdanilowicz . Using the icon looks much cleaner. I have updated the PR with the changes.

Screenshot 2023-03-14 at 10 42 41 AM

packages/mirrorful/editor/src/components/ColorPalette/EditColorModal.tsx
```diff
@@ -81,21 +86,30 @@ export function EditColorModal({
...
-                }}
-              />
+                <Flex flex="1">
+                  <FormLabel>Variable Name</FormLabel>
+                  <Tooltip
+                  placement="right"
+                  closeDelay={500}
+                  hasArrow
+                  label={"Variable names don't need a hyphen."}>
+                  <InfoIcon css={{ marginTop: '5px', marginLeft: '-6px' }} />
```
alexdanilowicz, line 96: Ideally, we shouldn't use negative margin, instead we should use marginRight.

packages/mirrorful/editor/src/components/ColorPalette/EditColorModal.tsx
```diff
@@ -81,21 +86,30 @@ export function EditColorModal({
...
-                onFocus={(e) => {
-                  setShowBaseColorPicker(true)
-                }}
-                onKeyPress={(event) => {
-                  if (event.key === 'Enter' && baseRef.current) {
-                    baseRef.current.focus()
-                  }
-                }}
-              />
+                <Flex flex="1">
```
alexdanilowicz, line 89: Is flex=1 needed here?

packages/mirrorful/editor/src/components/ColorPalette/EditColorNameModal.tsx
```diff
@@ -50,7 +52,16 @@ export function EditColorNameModal({
...
           <FormControl>
-            <FormLabel>Color Name</FormLabel>
+            <Flex flex="1">
+              <FormLabel>Color Name</FormLabel>
+              <Tooltip
+                  placement="right"
+                  closeDelay={500}
+                  hasArrow
+                  label={"Variable names don't need a hyphen."}>
+                    <InfoIcon css={{ marginTop: '5px', marginLeft: '-6px' }} />
```
alexdanilowicz, line 62: Same comments as above, marginRight vs negative margin

packages/mirrorful/editor/src/components/ColorPalette/EditColorNameModal.tsx
```diff
@@ -50,7 +52,16 @@ export function EditColorNameModal({
           }}
         >
           <FormControl>
-            <FormLabel>Color Name</FormLabel>
+            <Flex flex="1">
```
alexdanilowicz, line 55: same comment about whether 1 is needed

packages/mirrorful/editor/src/components/Onboarding/pages/NamePrimary.tsx
```diff
@@ -113,12 +113,21 @@ export function NamePrimary({
...
+              Color Name:
+            </Text>
+            <Tooltip
+                  placement="right"
+                  closeDelay={500}
+                  hasArrow
+                  label={"Variable names don't need a hyphen."}>
+                <InfoIcon css={{ marginTop: '5px', marginLeft: '4px' }} />
+            </Tooltip>
+          </Flex>
```
alexdanilowicz, line 130: I think this is not needed here, it's too much on this page. I think also because the word is "Color Name" vs "Variable Name" its more clear. 

Screen Shot 2023-03-14 at 9 47 31 AM

packages/mirrorful/editor/src/components/ColorPalette/EditColorModal.tsx
arneesh: Removed it.

packages/mirrorful/editor/src/components/ColorPalette/EditColorNameModal.tsx
arneesh: Removed it.

packages/mirrorful/editor/src/components/ColorPalette/EditColorModal.tsx
arneesh: Hey @alexdanilowicz , yeah Ideally I would not have used negative margins. But I don't think that marginRight will work in this scenario due to the box structure of FormLabel field, it contains some inbuilt spacing and rightMargin won't overlap it. I can remove it all together but then the space between FormLabel and InfoIcon also increases. Had to use negative margin on purpose to reduce that spacing.

Screenshot 2023-03-14 at 11 00 56 PM

alexdanilowicz: oh that makes sense, thanks for teaching me something new and the detailed screenshot!

alexdanilowicz:
Awesome! Well done @arneesh. Going to merge this in. 🎊

Issue #113, opened 2023-03-08
Sort fontsizes by size 


alexdanilowicz:
In this example, `0.75 rem` should be first, even if it was added last.

For now, we can just sort by unit. For example, you can sort the rems, then px, then em. Then you can avoid crazy math.

<img width="913" alt="Screen Shot 2023-03-08 at 3 46 18 PM" src="https://user-images.githubusercontent.com/29822597/223878482-909b9519-bf7d-46b4-9efa-f38790e1c7a5.png">

joelcmk:
Hey Alex, I can help with this issue!

alexdanilowicz:
Resolved

PR #145, opened 2023-03-13
Remove leading zeros when deleting a variant value

joelcmk:
While I was on issue #113 I noticed that it was tricky to edit the variant value. Whenever you delete the current value it add a zero to the left. Here is an example: 
Screen Shot 2023-03-13 at 4 13 36 PM

With this update, the user would be able to edit the variant value without getting a zero at the left of the input field. It would look like this instead. 
Screen Shot 2023-03-13 at 4 16 02 PM


alexdanilowicz:
I'm totally aligned, I noticed this too. Thank you so much for pushing this up. Just confirming @Teddarific you did not make the default 0 intentional? I found the leading 0 annoying as well, so @joelcmk I think this is great.

alexdanilowicz:
QA-ed. looks great. merging.

Issue #147, opened 2023-03-14
Add ability to select which file types Mirrorful exports


zachsnoek:
We should let users select which file types Mirrorful exports because some/most projects don't need tokens in all formats. We can enable this by adding a property like `files` to the configuration object whose value is an array of export file types.

See [this Slack thread](https://mirrorful.slack.com/archives/C04QRA1GA4C/p1678668833244879?thread_ts=1678663135.944999&cid=C04QRA1GA4C).

zachsnoek:
@alexdanilowicz / @Teddarific, feel free to assign this to me 🙋🏻

alexdanilowicz:
Assigned, love it. I myself have been wanting this. Thanks for opening an issue so that it is tracked!

PR #148, opened 2023-03-14
feat: add files property to configuration object

zachsnoek:
## Summary

This PR sets us up for letting users select which file types to export (#147). It adds a `files` property to the configuration object and a corresponding migration. In my next PR, we'll add an input in the editor for setting this property and only export to the specified files.

## Testing

Manual testing: Exported to a configuration file at the migration version `0.0.3` and ensured that the migration version was incremented and a `files` property was created.


alexdanilowicz:
Nice, I love how you kept the migration separate. Fantastic! I'll merge this in now.

Issue #155, opened 2023-03-14
Token objects generated for theme.js and theme.ts have unnecessary quotations around all property names 


alexdanilowicz:
Token objects generated for theme.js and theme.ts have unnecessary quotations around all property names because we just throw the JSON.stringified version in there.

## Note:

A note: prettier usually cleans this up automatically if you touch the file, which is nice, but I see the value of this issue (which came from [Slack](https://mirrorful.slack.com/archives/C04QRA1GA4C/p1678663135944999)). I will say I like the simplicity of JSON.stringify, and I don’t think it’s worth a bunch of code to maintain/add to the project just to strip out the quotes. But if anyone thinks the solution will be simple, go for it!

zachsnoek:
Hey, @alexdanilowicz –– you can assign this one to me. I have a change coming that should make this pretty straightforward.

Totally agree about the simplicity of `JSON.stringify` and maintenance cost. For nit-pickers like me, I think the extra polish  will be nice; having to manually run Prettier to strip the quotation marks is a little annoying 😅

PR #160, opened 2023-03-15
refactor: add file type translators

zachsnoek:
# Summary

Currently, two file generation functions in the export API handler are responsible for generating all six token files. While this works fine now, it'll become hard to navigate and maintain ~if~ when we start supporting more export types. The current implementation also makes it more difficult to generate specific file types.

Instead, we should have functions that handle the content translation for each file type. This PR introduces translator functions for the six file types and replaces the current handler implementation. These signatures are probably incomplete but should be enough to help us add new export types more quickly, and with my next PR (addressing #147), that will use the config to determine which files to generate.

# Notes

The easiest way to review this is by reviewing each commit.

I noticed that the current file generation functions are `async` but using `fs.writeFileSync`. I kept the `toContent` functions synchronous for now, but in the future, we might want to make them `async`, map them to Promises, then `Promise.all` them in the handler.

Lastly, these changes will help me address #155 more easily 😉.

# Testing

I exported my config manually and ensured the refactored code generated the same files. Please double-check and confirm that the output has stayed the same!

### DIFF
--- a/packages/mirrorful/editor/src/pages/api/export.ts
+++ b/packages/mirrorful/editor/src/pages/api/export.ts
@@ -1,96 +1,11 @@
 import { NextApiRequest, NextApiResponse } from 'next'
 import fs from 'fs'
-import { TColorData, TTokens } from 'types'
+import { TExportFileType, TTokens } from 'types'
 import { rootPath, store } from 'store/store'
-
-const sanitizeName = (name: string | number) => {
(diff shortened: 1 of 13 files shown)


Teddarific (approved):
This is... amazing, Zach! Thanks for cleaning this all up and making it extensible. Love it.

Teddarific:
QA'ed, and merging!

Issue #152, opened 2023-03-14
Add ability to manually reorder colors


alexdanilowicz:
As a user, my "main" color might be purple, but I added purple last to Mirrorful. As a result, purple looks like it's not important because it's at the bottom, but in reality I use it all the time. For example, think about Doordashes theme: red is probably the most important color. 

Add ability to manually reroder colors. I think ideally there is some sort of drag and drop functionality as that is the most intuitive, but probably also the most complicated. This is more of a polish ticket, and we don't want to bloat the codebase, so let's discuss this a bit here before getting started.

Teddarific:
Just to clarify, are you referring to re-ordering variants or colors?

alexdanilowicz:
This issue was created from @zachsnoek's thread (feel free to chime in Zach), but IMO: reorder colors

For example:
```
BLUE

PURPLE
```

But then I add green, which is my new primary. I would green to come first.

For variants, I would expect them to be sorted in order: 50-900.

karanvirsb:
Hey, can I be assigned to this issue?

PR #169, opened 2023-03-16
Ability to manually reorder variants #152

karanvirsb:
This is a solution for #152 but there is a big underlying issue. We need each colour to have a unique id. So TColorData needs an id property for the drag and drop to work correctly. But as of right now, it works without animation.


Teddarific:
Hey @karanvirsb, this is looking really promising so far! We just implemented a bit of a redesign, so apologies if they mess up your branch here. Happy to help get things sorted out with merge conflicts and stuff — let me know if you need any help! (Also feel free to directly ping me on Slack!)

karanvirsb:
I resolved the conflicts but will need to fix the ColorDisplay Merge as I need to add drag functionality to that.

Teddarific:
This is looking sweet so far Karan, let me know when it's ready to review again!

karanvirsb:
@Teddarific thank you, it can be reviewed now for any feedback!

Teddarific:
Hey @karanvirsb! Just checked out the branch, it overall looks sweet, you definitely got an eye for UI! 

I think we unfortunately might need to hold off merging this for now. I think with the new dashboard design, the drag & drop becomes tougher in the case where there are a lot of variants (e.g. I had a hard time dragging and dropping with a smaller screen + with the default variants).

Going to pause this for now, but this is definitely something we will come revisit since it's definitely something we want — just need to think about how we want to allow people to reorder!  Thanks for trailblazing the UI here!!

karanvirsb:
> Hey @karanvirsb! Just checked out the branch, it overall looks sweet, you definitely got an eye for UI!
> 
> I think we unfortunately might need to hold off merging this for now. I think with the new dashboard design, the drag & drop becomes tougher in the case where there are a lot of variants (e.g. I had a hard time dragging and dropping with a smaller screen + with the default variants).
> 
> Going to pause this for now, but this is definitely something we will come revisit since it's definitely something we want — just need to think about how we want to allow people to reorder! Thanks for trailblazing the UI here!!

Thanks for the feedback! I can agree with the point that we need to fix the layout a bit and maybe afterwards we can decide how users can reorder the colours.

Teddarific:
@karanvirsb Glad you agree! Let me keep this in mind, and you'll be the first person I ping when I think we're ready to implement reordering :) Thanks again for your hard work here, even though the code didn't make it in, its SUPER DUPER valuable seeing how we would do this in the future!

Going to close this for now, but this will be a PR to reference in the future!!

PR #171, opened 2023-03-17
feat: add ability to select export files

### LINKED ISSUE #147: Add ability to select which file types Mirrorful exports

zachsnoek:
# Summary

This PR resolves #147: It lets users select only the file types they want to be exported!

[Demo](https://drive.google.com/file/d/1j1gGfNc1pCTTrLyHK7gCDJcX274Nk_2T/view?usp=share_link) (Ugh, sorry for the Google Drive link.)

**UI changes**:
* Added an export settings modal that has a file type section for selecting export files
* Added a settings button to open the export settings modal
* Added a step to the onboarding flow for selecting file types so that the user starts with only the files they need

**API changes**:
* `/api/config` now returns the config object, not just the tokens
* `/api/export` now accepts the config object, not just the tokens

# Testing

* Manually tested creating a new Mirrorful config from scratch (see demo video above) and exporting tokens
* Manually tested updating file types with an existing Mirrorful config and exporting tokens: only the selected file types should get updated, and the other files should remain untouched

# Notes

* Note that currently, Mirrorful won't delete token files if the user removes an export file type, so existing projects can consider manually deleting the unneeded files if they remove export file types.
* Obligatory: This is primarily a duct tape implementation. I see a lot of opportunities for refactoring (e.g., updating state management and creating reusable components), but this is good enough for now™️.

### DIFF
--- a/packages/mirrorful/editor/src/components/Dashboard.tsx
+++ b/packages/mirrorful/editor/src/components/Dashboard.tsx
@@ -50,8 +50,7 @@ export function Dashboard() {
     await fetch('/api/export', {
       method: 'POST',
       body: JSON.stringify({
-        colorData: colors,
-        typography,
+        tokens: { colorData: colors, typography },
       }),
     })
 
@@ -63,8 +62,10 @@ export function Dashboard() {
     await fetch('/api/export', {
       method: 'POST',
       body: JSON.stringify({
-        typography,
-        colorData: data,
+        tokens: {
+          typography,
+          colorData: data,
+        },
       }),
     })
   }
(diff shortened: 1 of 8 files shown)


Teddarific:
Sweet, thanks @zachsnoek !

Issue #170, opened 2023-03-17
Font preview size limitation to avoid page structure issues


gravityNotWorking:
If we add a new font and assign it an absurd variant value, it could potentially cause the page structure to break. To prevent this, it would be great to set a maximum size for the font previews. One possible solution could be to limit the previews to a size of no more than 8rem.

![image](https://user-images.githubusercontent.com/77449139/225789043-718172c8-8ba7-4ffb-ad03-2bd48ba91d37.png)
![image](https://user-images.githubusercontent.com/77449139/225789033-3496b905-cea9-4588-87ab-22530c277ff4.png)

PR #173, opened 2023-03-17
fix: max font size in preview

Saganic:
Fixes: #170 

- Adds a function that limits the fontSize to 6rem, 6em and 96px for the FontSizeRow component preview


alexdanilowicz:
Looks good to me! Merging.

Teddarific:
Hey, thanks for the PR here! 

Can we actually update this logic to show like "FONT SIZE TOO LARGE TO BE RENDERED" instead of the text example? Otherwise, I think it's misleading to show a different font size than what is entered. For example, if I enter `8rem`, I might think "oh this is what "8rem" looks like when in reality I'm looking at `6rem`

PR #175, opened 2023-03-17
refactor: show font size too large message in FontSizeRow

Saganic:
- Refactor check for whether the fontSize is too large to return a boolean
- Show "Font size {value}{unit} too large to render." message when the fontSize is too large for FontSizeRow preview


Teddarific:
LGTM! Thanks, Chris!

PR #179, opened 2023-03-18
sort color values by lightness

### LINKED ISSUE #176: Sort color variants
We should sort the color variants from lightest to darkest in `ColorDisplay`

image

A word of caution, sorting colors may not be as easy as you think it is! 

p.s. if a certain question comes up, probably the answer is "sort by hue". maybe try `tinycolor(color).toHsl()` and sort by the `h` value.

ktra99:
resolves #176 

Screenshot 2023-03-18 at 11 40 19

# Notes
This pr sorts the lightness of a variant and sets up a base case. Extreme cases like adding a completely different color to a variant hasn't been taking into account yet.


alexdanilowicz:
Awesome @ktra99 !

Issue #154, opened 2023-03-14
Add vs Edit language — Clicking “Add New Font Size” show a modal with the title “Edit Font Size Variant”.


alexdanilowicz:
Right now, the modal title says Edit when it should say Add (unless it already exists because then you are indeed "editing"). Should be a quick fix.

gravityNotWorking:
I would like to take this issue. Additionally, I noticed that when we click on the `Add New Font Size` button, in the popup "Delete" button is still displayed. This should not be the case if we are adding a new font size.

PR #180, opened 2023-03-18
Modal title fixed

gravityNotWorking:
Solves this issue https://github.com/Mirrorful/mirrorful/issues/154


alexdanilowicz:
Nice work! Made one change fyi @gravityNotWorking. There is no need to use `useState` because state does not change. We can simply pass true or false. Check out the commit here: https://github.com/Mirrorful/mirrorful/pull/180/commits/5f651236c49f97200685015cda120d5e55108c02

gravityNotWorking:
That makes more sense, actually!

PR #182, opened 2023-03-18
Updated color selection functionality with lookup approach, fixes #58

### LINKED ISSUE #58: bug: Color labeled as unknown

prichter:
** Summary **
Previously implementation of color naming often had unknown appearing.  When this was resolved, there were inaccuracies in the color / hue descriptions being calculated.  This is a different approach using a curated list of colors, from which the closest matching HEX RGB color is selected and the associated name returned.

** Details **
In order to get around hue description mismatches, this PR adds a list of curated colors taken from `https://chir.ag/projects/ntc/ntc.js`, and uses an algorithm to grab the nearest color name based on diff matching of the incoming color RGB /HSL values and the values in the list.  This does add a significant amount of data into the utils file, but the payoff is increased accuracy of naming the colors.  Inspired by creative commons 2.5 licensed code from Chirag Mehta.


prichter:
image
This is how it looks.

alexdanilowicz:
Interesting! Thanks for this Paul. I have to say, it is a bit weird to see "dark purple" but the name is "Jacarta." I've also never heard of "Finn", "Fern Frond", or a few of the other color names on the list... 

Let me think on this one a bit and do some user testing. I LOVE that "unknown" is no longer a possibly, but I think the best user experience might be what we had a version of what originally, and then if it's unknown we use this list.

I say think because I think most colors are simply some variation of: red, blue, green, purple, orange, pink, yellow, brown, white, black, grey... and they should be named as such. For example at Liveramp, the color name was simply "Liveramp Green" not "Bonza Green". Most companies simply name the color the name of their company + the simple color name. And I just hope this doesn't introduce confusion. 

I like the approach tho. Let me get some input and thank you so much for opening the PR!

prichter:
@alexdanilowicz - I understand your thoughts re color naming, the reason I liked this list approach is that it helps obviate the "is this red or pink" question which can happen with a subjective analysis of hues.  This approach also allows an option to (re)define a reduced list of colors more conventionally named or even rename the colors in this list if desired - the algorithm would then select the nearest from the reduced / renamed list...  Just some food for thought.

alexdanilowicz:
> "is this red or pink" question which can happen with a subjective analysis of hues.

Very true.

alexdanilowicz:
Alright, I'm cool with merging this, but if we hear some confusion for users might tweak it a bit. Thanks for your hard work @prichter!

PR #185, opened 2023-03-19
Update styling of base color variant - fixes #107

### LINKED ISSUE #107: Add tooltip to variant square that is assigned as the base
If you add a color in the editor, it automatically creates a "500" variant, which when edited shows the "Set as base" flag set.  If you delete this _variant_, because of the flag set, it also deletes the base color.  This is not immediately obvious before the fact.  Suggest a "500 (BASE)" description in tile to make this clearer, or similar mechanism.

prichter:
Change styling of variant to highlight if it matches base - bolder font, thicker border, text indicating 'base'.  Update base comparison so that all colors are upper-cased for comparison.  Proposed fix for issue #107.


alexdanilowicz:
Attaching a screenshot. Let me see what @Teddarific thinks:

image

Teddarific:
LGTM!

PR #186, opened 2023-03-19
Restore 'Add Color' button, as skeleton no longer matched

prichter:
Simplified skeleton to 'Add Color' button, as styling was inconsistent with new layout.  Another issue can be raised to put placeholders in `AddColorSkeleton.tsx` later if desired.


alexdanilowicz:
Makes sense. I'm aligned.

Issue #116, opened 2023-03-09
If mirrorful is not running, kill localhost in your browser window


alexdanilowicz:
As a user, if I kill mirrorful in my terminal, but then go to my browser, I might be confused mirrorful is still rerunning. We should send a command that it's no longer running.

alexdanilowicz:
It was thrown out as socket.io being an option potentially. Might be a little heavy handed, though.

alexdanilowicz:
Closing this because not many users have reported this since we changed to new package system and I think the long polling might lead to issues down the road.

PR #197, opened 2023-03-20
Checking if mirrorful server has been shutdown (includes prettier changes)

karanvirsb:
This is a solution to issue #116. It uses long polling to check if the server is up. If the server is closed then the cli sends a fetch to an endpoint that sends it back to the polls.


packages/mirrorful/package.json
```diff
@@ -48,6 +48,7 @@
     "cross-spawn": "^7.0.3",
     "framer-motion": "^9.0.4",
     "fs": "^0.0.1-security",
+    "mirrorful": "^5.0.0",
```
alexdanilowicz, line 51: I don't think this should be here, I presume this was added during testing?

packages/mirrorful/editor/src/pages/index.tsx
```diff
@@ -1,8 +1,38 @@
...
 import { TConfig } from '@mirrorful/core/lib/types'
 import Head from 'next/head'
+import { useState, useRef, useEffect } from 'react'
 
 export default function Editor() {
+  const [hasShutDown, setHasShutDown] = useState(false)
+  const counter = useRef<boolean>(false)
+  useEffect(() => {
+    function pollForServerEndCheck() {
+      fetch('http://localhost:5050/api/longPoll', { keepalive: true })
```
alexdanilowicz, line 12: Hmm.. the only tricky thing is in the future we wanted to allow the user to pass their own port, and so hardcoding to be 5050 I'm not sure about

packages/mirrorful/editor/src/pages/index.tsx
```diff
@@ -1,8 +1,38 @@
...
+        .then((res) => res.text())
+        .then((res) => {
+          if (res === 'exiting') {
+            setHasShutDown(() => true)
```
alexdanilowicz, line 16:
```suggestion
            setHasShutDown(true)
```

packages/mirrorful/editor/src/pages/index.tsx
```diff
@@ -1,8 +1,38 @@
...
+            counter.current = true
+          }
+        })
+        .catch(() => {
+          if (navigator.onLine) {
+            setHasShutDown(() => true)
```
alexdanilowicz, line 22:
```suggestion
            setHasShutDown(true)
```

packages/mirrorful/editor/src/pages/index.tsx
```diff
@@ -1,8 +1,38 @@
 import { Dashboard } from '@mirrorful/core/lib/components/Dashboard'
+import ServerEndedMessage from '@mirrorful/core/lib/components/ServerEndedMessage'
 import { TConfig } from '@mirrorful/core/lib/types'
 import Head from 'next/head'
+import { useState, useRef, useEffect } from 'react'
 
 export default function Editor() {
+  const [hasShutDown, setHasShutDown] = useState(false)
+  const counter = useRef<boolean>(false)
```
alexdanilowicz, line 9: What's the reason for the `counter`? It also doesn't make a ton of sense to me that it's a Ref? I feel like it should just be state? I'm curious though!

packages/mirrorful/editor/src/pages/api/longPoll.ts
```diff
@@ -0,0 +1,10 @@
+import type { NextApiRequest, NextApiResponse } from 'next'
+
+// to handle the requests
+import pool from '../../utils/connections'
+
+export default function handler(req: NextApiRequest, res: NextApiResponse) {
+  res.setHeader('Transfer-Encoding', 'chunked')
+  res.setHeader('Content-Type', 'text/html')
+  pool.push(res)
```
alexdanilowicz, line 9: I'm a bit confused, is `pool` just an empty array? Does it need to be imported? Or can we just do a res.send?

alexdanilowicz (commented):
Good start! A few comments and things I'd love to learn or have clarified when you get a second.

packages/mirrorful/editor/src/utils/connections.ts
```diff
@@ -0,0 +1,15 @@
...
+// export function sendEndingMessage() {
+//   console.log(connections)
+//   connections.map((res, i) => {
+//     console.log(i, 'ending')
+//     res.write('exiting')
+//     res.end()
+//   })
+// }
+
+export default connections
```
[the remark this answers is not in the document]
alexdanilowicz: Like to me, all this file does is export out an `[]` of type `NextApiResponse`

packages/mirrorful/editor/src/pages/index.tsx
karanvirsb: Yeah, we can make it so that it accepts a port in the future. But we would need to know where that variable for the port is located.

packages/mirrorful/editor/src/pages/index.tsx
karanvirsb: Yeah, the counter needs to be changed to a different name. But essentially in testing useState did not work by itself it needed a ref. I can try again soon to see if it works now!

packages/mirrorful/editor/src/pages/api/longPoll.ts
karanvirsb: The pool contains all the requests. What I was trying to do was when the server received the 'sigterm'  or 'sigint', it would loop over the pool and send a message back saying exiting. But since the server closes too fast, the pool just stores the request. We could potentially remove the pool but we cannot send a response back as that would end the long poll.

karanvirsb: Removed pool as it is not needed!

packages/mirrorful/editor/src/pages/index.tsx
karanvirsb: Made Url and port into their own variable for now.

yarn.lock
```diff
@@ -1073,9 +1073,32 @@
     "@jridgewell/resolve-uri" "^3.0.3"
     "@jridgewell/sourcemap-codec" "^1.4.10"
 
+<<<<<<< HEAD
```
alexdanilowicz, line 1076: Merge conflict in this file FYI

alexdanilowicz:
Hey @karanvirsb thanks for all your hard work here. I'm a bit worried a long polling solution might lead to a lot of overheard maintenance for little gain (because it's not THAT often the user accidentally kills it), but will take a peek when I get a second. Thank you so much!

karanvirsb:
> Hey @karanvirsb thanks for all your hard work here. I'm a bit worried a long polling solution might lead to a lot of overheard maintenance for little gain (because it's not THAT often the user accidentally kills it), but will take a peek when I get a second. Thank you so much!

Yeah, no problem. If it is too much overhead we can through this idea out. But long polling is better than using sockets which would ping every few seconds. I did look into server-sent events but it does not work as well due to the server closing and not being able to emit the event in time.

packages/eslint-plugin/rules/no-hardcoded-colors.md
```diff
@@ -25,15 +25,15 @@ const color = 'rgb(0, 0, 0)';
 Examples of **correct** code for this rule:
 
 ```js
-<Button
+;<Button
```
alexdanilowicz, line 28:
```suggestion
<Button
```

alexdanilowicz:
Hey @karanvirsb. getting back to this one because i actually really like this feature and want to merge it in:

1. i resolved merge conflicts for you. FYI looks like lot of changes, but its really just prettier changes in the examples dir, so we can ignore it.

2. not sure it works, does it work for you? i got this "error"

Screen Shot 2023-05-01 at 5 09 35 PM

3. we'll have to add a thing that only does the long polling on the package and not on web.

alexdanilowicz:
@karanvirsb when you're done working on this lmk and happy to merge it in. feel free to post in #contributors

packages/mirrorful/editor/src/pages/_app.tsx
```diff
@@ -127,7 +127,7 @@ export default function MyApp({ Component, pageProps }: AppProps) {
           }
         })
     }
-    if (!isShuttingDown.current) {
+    if (!isShuttingDown.current && process.env.NODE_ENV !== 'development') {
```
alexdanilowicz, line 130: instead, let's change dev to run on 5050 and remove this check

alexdanilowicz:
Closing this because we are going to focus on building out the web version (not the localhost version) instead. Thanks so much for your hard work on this! The localhost version is just really tricky. 

Full speed ahead on web! 🎊

PR #198, opened 2023-03-20
Scaffolding api for custom theme presets

### LINKED ISSUE #196: Add preset themes function
We can think about how to integrate this into the UI later, for now, we should add a function that generates a preset color palette for popular developer tool or consumer companies. For example:

1. Supabase theme
2. Linear theme
3. Facebook theme
4. Doordash theme
5. Spotify theme 
6. Notion theme

... could also do themes like "tropical" and "sleek" etc

What I'm thinking here is a function that given a `preset` arg of type `supabse | linear | facebook | tropical ...` etc returns a json object that is the colors associated with that company and matches what is consumed in the Mirrorful store.json. (That's the high level idea.)

The idea here is a user can specify a company or preset theme, and then immediately consume it.

ktra99:
resolves #196

# Summary

This PR proposes a solution by exposing an API endpoint. The idea is to have the user choose a theme based on the front-end. Once the user has selected a theme, this API can be called to dynamically generate the proposed theme based on the name, directly to the `.mirrorful` folder. 

# Notes

The functionality is based on already implemented utility functions. I scaffolded everything together into an API endpoint for testing purposes. The API is set up in a way that can be consumed through a get request but can be expanded to be more dynamic later on. The themes have been hardcoded into the API as well with the current structure -

```
const themes = [
  {
    name: 'supabase',
    primary: '#3bae7a',
  },
  {
    name: 'linear',
    primary: '#5E6AD2',
  },
  {
    name: 'facebook',
    primary: '#4267B2',
  },
  {
    name: 'doordash',
    primary: '#FF3008',
  },
  {
    name: 'spotify',
    primary: '#1BD760',
  },
  {
    name: 'notion',
    primary: '#FFFDF9',
  },
]
```

Loom - https://www.loom.com/share/506e01de849a45c1a8396e3ea5f350b5

### DIFF
--- a/packages/mirrorful/editor/src/pages/api/preset/[name].ts
+++ b/packages/mirrorful/editor/src/pages/api/preset/[name].ts
@@ -1,0 +1,120 @@
+import { NextApiRequest, NextApiResponse } from 'next'
+import fs from 'fs'
+import { TConfig, TExportFileType } from '@mirrorful/core/lib/types'
+import { rootPath, store } from '../../../store/store'
+import { translators } from '@mirrorful/core/lib/translators'
+import { defaultTypography } from '@mirrorful/core/lib/types'
+import tinycolor from 'tinycolor2'
+/*
+you may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
+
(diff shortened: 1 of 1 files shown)


alexdanilowicz:
Talked to Kenny offline. Going to hold this one for now as we're likely going to use OpenAI's API to allow users to pass in any theme name. Great work Kenny! 🚀 🇸🇪

Issue #41, opened 2023-02-23
`create-react-app` Relative imports outside of `src/` are not supported


Teddarific:
![Screen Shot 2023-02-23 at 10 26 05 AM](https://user-images.githubusercontent.com/16343600/220998731-2dd5ea40-6220-4bf0-9475-4aa9b892283c.jpeg)

Teddarific:
Current recommended workaround: Manually make a copy of the `.mirrorful` folder in the `src` directory.

alexdanilowicz:
We plop .mirrorful in `src/` if that folder exists?

joelcmk:
I may be able to help with this issue. Does the folder `.mirrorful` have to be on `src` just for `create-react-app` or it applies to `nextjs`, etc. ?

alexdanilowicz:
Hey @joelcmk, Nextjs + Mirrorful works fine without the `.mirrorful` folder being in src. For example, see: https://github.com/Mirrorful/mirrorful/tree/main/examples/tailwind-next

That said, I don't think there's anything _wrong_ with it being in src in an NextJs app, and it might make more sense in src anyway.

At the same time, I don't think the core problem where mirrorful MUST be in src is limited to just CRA (create-react-app). 

Let me know if any other questions. Do you have any ideas on how you would want to implement it? Some options:

1. Prompt the user when they run `npx mirrorful` for the first time and let them choose the directory of their choice (suggesting src dir). If .mirrorful already exists, don't prompt them
2. Plop it in root to start as we do today (the .mirrorful folder is created[ when the script runs here](https://github.com/Mirrorful/mirrorful/blob/main/packages/mirrorful/cli/init.ts#L31)) and then allow the user in the export config modal to choose where to put it. (In practice, I think this might be confusing to users' as it's not their job to know where the .mirrorful should live)
3. If `src` exists, put it in src.

@Teddarific any thoughts? Excited for your interest in @joelcmk. Let us know if any thoughts on questions 🎸

Teddarific:
Let's just always plop it in `src` if it exists for now.

Ideally in the long run its a user-controlled setting where they want it to export.

An interesting idea could be every time it exports, it looks for a `.mirrorful` folder so if a user moved it, it would automatically adjust to where the current `.mirrorful` folder is

alexdanilowicz:
As we are moving towards a web-based editor (Mirrorful Cloud), going to sunset this issue for now, but thank you for all those that helped!

PR #200, opened 2023-03-20
Fixes issue #41

joelcmk:
With this new code, if there's an `src` folder it will create a `.mirrorful` repository inside of it. Otherwise, it would do the same as before. I'm sorry for the delay, I had never done something like this. It took me a while to understand the issue but I learned a lot.


alexdanilowicz:
Hey @joelcmk — thanks again for this. I'll review this and merge this in later today.

PR #203, opened 2023-03-21
Fixed duplicate sorting

karanvirsb:
Just updated the sorting variants to remove duplicates from the sort. 

before

After the change:

updated


Teddarific:
Hey @karanvirsb! We actually want to show duplicates because we shouldn't stop users from adding duplicate values. It would be confusing to add a new variant, and for it to now show up because it's a duplicate. Let me know if you have any questions!

Issue #192, opened 2023-03-20
Add `build` Github action for `packages/mirrorful`


Teddarific:
Let's add Github actions that compile each package on PR to ensure that `main` is always in a pristine state.

Saganic:
I'll take this.

Teddarific:
Sweeeeet @Saganic, thanks for taking this one on!! Adding some additional color:

- Vercel covers us for web, and we don't actually super care about `packages/core` for now, so let's just start with one Github action for `packages/mirrorful`
- It might be a little bit tricky, because `packages/mirrorful` has `@mirrorful/core` as a dependency, and ideally we read off the local `@mirrorful/core` instead of pulling from NPM. Might need to do try to build `@mirrorful/core`, `yarn link`, and then build `packages/mirrorful` if that makes sense

PR #207, opened 2023-03-21
feat: add github action build

Saganic:
Fixes: #192 

- adds build workflow for packages
- installs and builds /packages/core
- links /packges/core
- instsalls and builds /packages/mirrorful

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,41 @@
+name: build
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  build:
+    name: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,41 @@
+name: build
```
Teddarific, line 1:
```suggestion
name: Package
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,41 @@
...
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  build:
+    name: build
```
Teddarific, line 9:
```suggestion
    name: Build
```

Teddarific (approved):
amazing dude!!

PR #208, opened 2023-03-21
Preset theme

karanvirsb:
Still adding more themes but I wanted some feedback if it's in the right direction of #196. The API endpoint could be taken away if not needed.


alexdanilowicz:
Going to close this in favor of #198. But I think your hardcoded themes are _very_ helpful and we will likely end up using them. We just want to integrate Open AI's API to allow the user to autogenerate it entirely.

PR #209, opened 2023-03-21
Support src directories 

joelcmk:
I had to reopen this PR since I closed the other one by mistake. 😢

### DIFF
--- a/packages/mirrorful/cli/init.ts
+++ b/packages/mirrorful/cli/init.ts
@@ -29,8 +29,6 @@ export async function init({
     process.exit(1)
   }
 
-  await makeDir('.mirrorful')
-
   const port = 5050 // don't hard code this
   // await openBrowser({ url: 'http://localhost', port })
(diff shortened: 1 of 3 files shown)


alexdanilowicz:
Hey @joelcmk, thanks for making this, so I think the path the API uses would need to be changed along with the CLI. I can take a look down the road. 

Today the way it works:
- Mirrorful CLI creates the .mirrorful folder (hardcoded to be at root)
- Mirrorful API exports files into the .mirrorful folder (hardcoded to be at root)

With this change, as-is:
- Mirrorful CLI creates a .mirrorful either in src or not
- Mirrorful API still exports files in to the .mirrorful folder (hardcoded to be at root)

So need to tweak this a bit, potentially save where the .mirrorful file is in the store.json

alexdanilowicz:
Going to mark this as draft, and happy to close it if you're not longer interested in working on it.

PR #215, opened 2023-03-22
Typography section overlap fix

### LINKED ISSUE #181: Typography title is overlapping with the sidebar on the small screen.
I know that the app is not currently mobile responsive, but even on my 1440x900 screen, the title and second title are overlapping with the sidebar, as shown.

image

gravityNotWorking:
I'm not sure this is the best practice but I checked Colors section too, both works same. Fixes #181 

Recording 2023-03-22 at 13 34 41


alexdanilowicz:
Makes sense to me! Responsive layouts are always tricky. Thanks for the Loom. Excellent work!

PR #217, opened 2023-03-23
Fix navbar issue

alexdanilowicz:
https://www.loom.com/share/e4ada435d0504d7f9af8af07c293f7aa

I think underlying issue is usage of width vs minwidth

### DIFF
--- a/packages/mirrorful/editor/yarn.lock
+++ b/packages/mirrorful/editor/yarn.lock
@@ -1279,6 +1279,9 @@
     tinycolor2 "^1.6.0"
     typescript "^4.9.5"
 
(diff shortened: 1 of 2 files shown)


packages/mirrorful/editor/yarn.lock
```diff
@@ -1279,6 +1279,9 @@
     tinycolor2 "^1.6.0"
     typescript "^4.9.5"
 
+"@mirrorful/eslint-config@file:../../eslint-config":
+  version "1.0.0"
```
alexdanilowicz, line 1283: not sure why i see this change, but that's what I get when I run yarn

PR #233, opened 2023-03-25
Fixing package.json for windows

karanvirsb:
Added fix for windows as core package json would give issues due to the cp command not existing on windows. Along with this change added new scripts to packages/mirrorful and removed removeFileOrDirectory.ts as it is no longer needed.


alexdanilowicz:
Oh very interesting, have not heard of https://www.npmjs.com/package/run-script-os before.

I think `run-script-os` should be a dev dependency tho, and installed with `npm install --save-dev run-script-os`.

Looks like you have it installed a regular dependency. Could you fix that @karanvirsb  ? Thanks for opening this!

karanvirsb:
> Oh very interesting, have not heard of https://www.npmjs.com/package/run-script-os before.
> 
> I think `run-script-os` should be a dev dependency tho, and installed with `yard add -D run-script-os`.
> 
> Looks like you have it installed a regular dependency. Could you fix that @karanvirsb ? Thanks for opening this!

I have fixed it and moved it into the dev dependencies!

alexdanilowicz:
Great, we'll get this merged in today!

Issue #222, opened 2023-03-23
Expand export preview to be more than just colors. Should also show typography and shadow example


alexdanilowicz:
Right now, in the export preview, all we show are colors. We should show how to import typography and shadows as well.

While we are here, might be also good to update the docs in `/docs`

<img width="818" alt="Screen Shot 2023-03-23 at 12 37 02 PM" src="https://user-images.githubusercontent.com/29822597/227331490-ff976397-f14a-403f-81bb-a90a0970eb84.png">

karanvirsb:
Hey @alexdanilowicz can I be assigned to this issue?

alexdanilowicz:
Sure go for it! @karanvirsb 🥳

PR #238, opened 2023-03-25
Add Typography and Shadows to exports

karanvirsb:
Added typography and shadows to exports. Still need to the docs! #222


packages/core/src/components/ExportSuccessModal.tsx
```diff
@@ -30,17 +30,129 @@ import { FiCheckCircle } from 'react-icons/fi'
...
+  return <DisplayTabs {...props}></DisplayTabs>
+}
+
+function TypographyTab(props: displayTabsProps) {
+  return <DisplayTabs {...props}></DisplayTabs>
+}
+
+function ShadowTab(props: displayTabsProps) {
+  return <DisplayTabs {...props}></DisplayTabs>
+}
```
alexdanilowicz, line 128: These are all the same component/code, so I think for sake of simplicity we can just have one and generalize the name as  function `TokenTab` ?

packages/core/src/components/ExportSuccessModal.tsx
```diff
@@ -30,17 +30,129 @@ import { FiCheckCircle } from 'react-icons/fi'
...
+type displayTabsProps = {
+  tabs: string[]
+  primaryName: string
+  cssPropertyName: string
+  cssName: string
+  javascriptPropertyName: string
+  javascriptName: string
+  tailwindPropertyName: string
+  tailwindName: string
```
alexdanilowicz, line 137: could we make these more type safe? For example, `muffin` is not a valid cssName or javascriptName or tailwind name, so I think the type should be more type safe. but instead of doing _every_ one (that would be ridiculous) lets just make it the values that we already use, ie, `tailwindName: "boxShadows" | "fontSizes"  ...` etc

docs/home/export-formats.mdx
```diff
@@ -10,7 +10,9 @@ We currently export to the following file types: `.js`, `.ts`, `.css`, `.scss`,
 If you are using Mirrorful's web editor, you will need to create the imported files yourself. You can do this by simply creating a file of your desired format (i.e. `.js`, `.css`, etc) and then copying
 the generated code into the file.
 
-## Using CSS
+## Colors
```
alexdanilowicz, line 13: Great work, I'm thinking maybe Colors should be it's own page... instead of export formats being this massive file. We use Mintifly to set up docs btw: https://mintlify.com/docs/introduction

No worries if you don't want to make it's own page, just an idea.

alexdanilowicz (commented):
Leaving some comments! Great work! Really nice, hope the comments are helpful.

alexdanilowicz:
Really nice @karanvirsb. Will look at this and try and get this merged in today.

alexdanilowicz:
Double tabs is a bit unconventional to me. Need to think on exact UI here

Screen Shot 2023-03-28 at 7 45 06 PM

packages/core/src/components/Dashboard/index.tsx
```diff
@@ -202,7 +202,7 @@ export function Dashboard({
       </Box>
       <ExportSuccessModal
         platform={platform}
-        primaryName={colors && colors[0] ? colors[0].name : 'primary'}
+        // primaryName={colors && colors[0] ? colors[0].name : 'primary'}
```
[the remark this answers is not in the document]
karanvirsb: Hey Mar Tuico thanks for the suggestion, but Dashboard technically will not be used anymore due to the next js routing. I will still accept it but we could potentially just delete it too.

alexdanilowicz:
@karanvirsb looks awesome to me! Really well done!!! Reviewing now. Fantastic! 🥇 

Screen Shot 2023-03-31 at 1 31 05 PM

PR #239, opened 2023-03-25
refactor: use JSON5 to stringify theme object

### LINKED ISSUE #155: Token objects generated for theme.js and theme.ts have unnecessary quotations around all property names

zachsnoek:
Resolves #155.

---

This PR replaces `JSON.stringify` with `JSON5.stringify` so that property names in `.cjs`, `.js`, and `.ts` theme files only have quotations when necessary.

`theme.js` with `JSON.stringify`:

Screenshot 2023-03-25 at 09 58 38

`theme.js` with `JSON5.stringify`:

Screenshot 2023-03-25 at 09 59 47

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -25,6 +25,7 @@
     "@emotion/styled": "^11.10.6",
     "@hello-pangea/color-picker": "^3.2.2",
     "framer-motion": "^10.6.0",
+    "json5": "^2.2.3",
     "lottie-react": "^2.4.0",
     "posthog-js": "^1.51.3",
     "react": "^18.2.0",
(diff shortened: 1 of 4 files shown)


Teddarific:
LGTM, thanks @zachsnoek! TIL about JSON5

Issue #241, opened 2023-03-26
Clicking on color value automatically copies and paste it to clipboard


Teddarific:
<img width="124" alt="image" src="https://user-images.githubusercontent.com/16343600/227808911-9f88f149-3774-42b5-b47f-52a3e3b255d8.png">

For example, clicking on "#b59be9" should automatically c/p to my clipboard. 

Ideally, we add a basic hover state to indicate that the value is clickable.

CC @ktra99 for the idea!

joelcmk:
Would it make sense to add this feature for the shadow boxes too?

alexdanilowicz:
I think so @joelcmk! Maybe you can add that as a fast follow after your [PR](https://github.com/Mirrorful/mirrorful/pull/248) is merged

PR #243, opened 2023-03-27
Add copy color feature

karanvirsb:
Created the copied feature #241. Here is a loom of the changes https://www.loom.com/share/e02d326fa4884ec195252a48289c493a

### DIFF
--- a/packages/core/src/components/ColorPalette/ColorDisplay.tsx
+++ b/packages/core/src/components/ColorPalette/ColorDisplay.tsx
@@ -2,6 +2,7 @@ import { DeleteIcon, EditIcon } from '@chakra-ui/icons'
 import {
   Box,
   Button,
+  Flex,
   Icon,
   IconButton,
   Menu,
@@ -83,7 +84,12 @@ function VariantRow({
         css={{ display: 'flex', alignItems: 'center', position: 'relative' }}
       >
         {hasCopiedHexCode ? (
-          <Text css={{ marginRight: '5px' }}>Copied -&gt;</Text>
+          <Text
+            fontSize="1rem"
+            css={{ marginRight: '5px', fontWeight: 'bold' }}
+          >
+            Copied &rarr;
+          </Text>
         ) : null}
         <Text
           fontSize="1rem"
--- a/packages/core/src/components/ColorPalette/ColorDisplay.tsx
+++ b/packages/core/src/components/ColorPalette/ColorDisplay.tsx
@@ -2,7 +2,6 @@ import { DeleteIcon, EditIcon } from '@chakra-ui/icons'
 import {
   Box,
   Button,
-  Flex,
   Icon,
   IconButton,
   Menu,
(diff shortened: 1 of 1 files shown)


packages/core/src/components/ColorPalette/ColorDisplay.tsx
```diff
@@ -2,6 +2,7 @@ import { DeleteIcon, EditIcon } from '@chakra-ui/icons'
 import {
   Box,
   Button,
+  Flex,
```
alexdanilowicz, line 5: Let's remove unused imports!

packages/core/src/components/ColorPalette/ColorDisplay.tsx
```diff
@@ -49,20 +64,51 @@ function VariantRow({
         padding: '0px 24px',
         borderRadius: 8,
         border: variant.isBase ? '2px solid black' : '1px solid black',
+        position: 'relative',
       }}
+      role="group"
+      _hover={{
+        cursor: 'pointer',
+        backgroundColor: tinycolor(variant.color).isDark() ? 'white' : 'black',
```
alexdanilowicz, line 72: I actually found the fact that is hovers to a different color (white) in this instance a it confusing. Because it made me think the color was changing to white. Maybe increase its border to be 2px instead?

packages/core/src/components/ColorPalette/ColorDisplay.tsx
```diff
@@ -99,6 +146,10 @@ function VariantRow({
       />
     </Box>
   )
+  function handleCopyColor(colorHex: string) {
+    navigator.clipboard.writeText(colorHex)
+    setHasCopiedHexCode(true)
```
alexdanilowicz, line 151: In the spirit of "colocation", let's just put this inline instead of separating it out to a function. It's only used in one place, so making its own function seems overkill.

packages/core/src/components/ColorPalette/ColorDisplay.tsx
karanvirsb: Will do!

packages/core/src/components/ColorPalette/ColorDisplay.tsx
karanvirsb: Ah okay I tried reversing the colors. I had tried the border out but maybe ill make it bigger.

packages/core/src/components/ColorPalette/ColorDisplay.tsx
```diff
@@ -49,20 +64,51 @@ function VariantRow({
...
     >
+      {hasCopiedHexCode ? (
+        <Text
+          fontSize="1rem"
+          css={{
+            marginRight: '5px',
+            fontWeight: 'bold',
+            position: 'absolute',
+            right: '100%',
+            zIndex: '100',
```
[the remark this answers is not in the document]
karanvirsb: It still works when the screen is shrunk but ill take a look at the tooltips might be better.

alexdanilowicz:
Screen Shot 2023-03-28 at 6 59 40 PM
Fantastic. Merging.

Issue #235, opened 2023-03-25
Add a box shadow generator 


joelcmk:
It would be nice if, instead of just having the option to add the values manually, there was a shadow generator where the user could change the values and have a preview of the result. I would be happy to implement this feature.

PR #248, opened 2023-03-28
Add box shadow generator

joelcmk:
Added a new feature that enables users to preview the box shadow. This enhancement improves the user experience and makes it easier to visualize how the box shadow settings impact the design. Thanks for reviewing the code and let me know if you have any feedback! Here is a screenshot!
Screen Shot 2023-03-27 at 11 17 37 PM
Screen Shot 2023-03-27 at 11 17 46 PM


alexdanilowicz:
Really cool Joel, FYI I think the build is failing because some lock files were committing (probably from running npm instead of yarn). I'm going to delete them and see if that fixes it!

alexdanilowicz:
Resolves #235

alexdanilowicz:
Oh I think there's some typescript errors @joelcmk . Maybe if you're free this weekend we could hop on a call and I can see if your editor is set correctly. Can you see the build errors if you click on the Vercel details above?
Screen Shot 2023-03-27 at 10 27 54 PM

joelcmk:
Yes, I can see them, I'll try to fix it later today. Also, thanks for letting me know about using npm vs yarn

joelcmk:
Hey Alex, the tests passed!

alexdanilowicz:
Looks awesome to me!!!

Screen Shot 2023-03-28 at 6 41 50 PM

packages/core/src/components/Shadows/ShadowColorPicker.tsx
```diff
@@ -0,0 +1,158 @@
...
+        blur: number
+        spread: number
+      }
+    | undefined
+}) {
+  const [color, setColor] = useState(
+    presetColor ? presetColor : 'rgba(1, 1, 1, 0.4)'
+  )
+
+  initialValues = initialValues
```
Teddarific, line 39: nit: While this works, probably shouldn't modify props. Can probably update the function signature instead to define a default value, e.g.

```
export function ShadowColorPicker({
   initialValues = { hOffset: 0, vOffset: 0, blur: 0, spread: 0 },
} : { 
  initialValues?: { hOffset: number, vOffset: number, blur: 0, spread: 0 }
}
```

alexdanilowicz: Good callout. Also, initial values should be an array, right? because you can have multiple values. For example:
Screen Shot 2023-03-28 at 7 21 48 PM

alexdanilowicz: Screen Shot 2023-03-28 at 7 22 32 PM
But appears to be a bug that only first value is captured

PR #255, opened 2023-03-29
Added hover effect on color copy

karanvirsb:
Added hover to hex codes.


Teddarific (approved):
Amazing work Karanvir, two small nits but otherwise looks fantastic!

Issue #253, opened 2023-03-29
Refreshing the page resets the current tab.


gravityNotWorking:
If you refresh the page while in the shadows or typography section, it will automatically return to the first tab which is colors section.

![Recording 2023-03-29 at 16 22 16](https://user-images.githubusercontent.com/77449139/228551759-714e21a8-66b2-4138-88bb-c71d2fc125a9.gif)

Teddarific:
@gravityNotWorking Yep, this would be nice to fix. I think we'll need to implement routing, which shouldn't be too bad!

Might be a bit of a refactor, but definitely worth it if anyone is interested in taking this one on!

https://nextjs.org/docs/routing/introduction

JoshuaRotimi:
@Teddarific @alexdanilowicz Can you assign this to me? I would like to work on it. Thanks.

Teddarific:
@JoshuaRotimi  this is actually fixed now with our latest change in #263! If you're looking for another good issue, maybe this one? https://github.com/Mirrorful/mirrorful/issues/270

alexdanilowicz:
@JoshuaRotimi also consider joining our slack! https://join.slack.com/t/mirrorful/shared_invite/zt-1ps2xtxh0-2NaixFfFzSKZbr5gw_AHfA

JoshuaRotimi:
Oh alright. I'll look at #270. I will be joining the slack as well. Thanks @alexdanilowicz @Teddarific

PR #263, opened 2023-03-30
Add nextjs routing

karanvirsb:
#253. Added next js routing functionality along with a splash screen to wait for loading.


alexdanilowicz:
Hey great work @karanvirsb! Need some time to review along with @Teddarific. Really appreciate all the hard work. You truly are Mirrorful's contributor of the week! I'm not familiar with [zustand](https://github.com/pmndrs/zustand) so need to study that a bit. I would expect using React context instead, but I see zustand argues there's less boilerplate. 

Do you have a particular reason you chose zustand or Context? 

Additionally, I think the sidebar is missing! At least on web:

Screen Shot 2023-03-31 at 1 27 08 PM

karanvirsb:
> Hey great work @karanvirsb! Need some time to review along with @Teddarific. Really appreciate all the hard work. You truly are Mirrorful's contributor of the week! I'm not familiar with [zustand](https://github.com/pmndrs/zustand) so need to study that a bit. I would expect using React context instead, but I see zustand argues there's less boilerplate.
> 
> Do you have a particular reason you chose zustand or Context?
> 
> Additionally, I think the sidebar is missing! At least on web:
> 
> Screen Shot 2023-03-31 at 1 27 08 PM

I choose zustand due to its simplicity in splitting up the pages into their slices. If it is too much I would love the feedback and can change it to context! Yes the web, I believe I know why it's due to hard coding the platform.

packages/mirrorful/editor/src/components/Layout.tsx
```diff
@@ -0,0 +1,136 @@
...
+import useMirrorfulStore from 'src/zustand/useMirrorfulStore'
+
+type props = { children: React.ReactNode; isLoading?: boolean }
+export function Layout({ children, isLoading = false }: props) {
+  const platform = 'package'
+  const router = useRouter()
+  const [tab, setTab] = useState<TTab>(
+    router.pathname === '/'
+      ? 'colors'
+      : (router.pathname.replace('/', '') as TTab)
```
alexdanilowicz, line 20: Hm, maybe instead of trying to get the path to be casted as a `TTab` we instead just change "/" to be a valid tab name. I.e., tab names and route names become the same thing.

Not a huge fan of the usage of "as"

packages/mirrorful/editor/src/pages/index.tsx
```diff
@@ -14,7 +32,13 @@ export default function Editor() {
         />
         <link rel="icon" href="/favicon.ico" />
       </Head>
-      <Dashboard
+      <Layout isLoading={isLoading}>
+        <ColorPaletteSection
+          colors={colors}
+          onUpdateColors={handleUpdateColors}
+        ></ColorPaletteSection>
```
alexdanilowicz, line 39: super nit, but I'm surprised the linter doesn't force us to make this just one tag, i.e.,

```javascript
        <ColorPaletteSection
          colors={colors}
          onUpdateColors={handleUpdateColors}
        />
```

packages/mirrorful/editor/src/pages/index.tsx
```diff
@@ -14,7 +32,13 @@ export default function Editor() {
...
+      </Layout>
+      {/* <Dashboard
```
alexdanilowicz, line 41: Let's just delete the commented out code.

alexdanilowicz (commented):
just some nits, will be curious what teddy thinks about zustand. very nice!

packages/mirrorful/editor/src/components/Layout.tsx
karanvirsb: I might keep this because the web edition still needs TTab. I will add a Todo comment to remove it!

packages/core/src/store/onBoardingSlice.ts
```diff
@@ -0,0 +1,21 @@
+import { StateCreator } from 'zustand'
```
Teddarific, line 1: I think it's a bit overkill bringing the onboarding logic in Zustand as well —

IMO the less information inside of Zustand the better — if it can be covered by a simple `useState`, it should be preferably done via a `useState` !

Probably the only data that needs to be in here is the design token data!

packages/mirrorful/editor/src/pages/_app.tsx
```diff
@@ -18,7 +24,49 @@ if (typeof window !== 'undefined') {
...
+      }
+
+      setColors(data.tokens.colorData ?? [])
+      setTypography(data.tokens.typography)
+      setShadows(data.tokens.shadows ?? defaultShadows)
+      setFileTypes(data.files)
+    } catch (e) {
+      // TODO: Handle error
+    } finally {
+      timeout.current = setTimeout(() => {
```
Teddarific, line 61: What is this timeout logic for?

packages/core/src/utils/fetchStoreData.ts
```diff
@@ -0,0 +1,8 @@
+import { TConfig } from '../types'
+
+export default async function fetchStoreData() {
+  const response = await fetch('/api/config')
```
Teddarific, line 4: This stuff should probably continue to live within the `packages/editor` context as it's only relevant to the package, and not for `web`. Trying to reserve `packages/core` for stuff relevant to both!

Teddarific (commented):
I think something top of my mind for me is how do we keep the Web <> Package stay in sync, and that adding stuff to the editor only requires one set of changes.

I'm worried that right now adding a new page requires adding a new file both to the package and the web version. Any ideas how we can do this? I wonder if routing can be done within a shared component.

packages/mirrorful/editor/src/pages/_app.tsx
karanvirsb: This timeout is used for the splash screen. So all data can be loaded in before using the app.

packages/core/src/utils/fetchStoreData.ts
karanvirsb: Okay understood will make the change!

packages/core/src/store/onBoardingSlice.ts
karanvirsb: I have removed onboarding from zustand. I agree it was kind of overkill.

packages/core/src/utils/fetchStoreData.ts
karanvirsb: Making this change means moving the core/src/component/Layout.tsx back to the editor. This change causes the sidebar to lose its collapsibility. As there seems to be a problem with its type definition.

packages/mirrorful/editor/src/components/Layout.tsx
```diff
@@ -0,0 +1,136 @@
+import { Box, Spinner, useDisclosure } from '@chakra-ui/react'
```
Teddarific, line 1: Intentional that `Layout.tsx` is duplicated in `packages/mirrorful` and `apps/web`?

Teddarific (commented):
This is looking great — let me know over Slack if there's stuff you're still working on, but tomorrow I'm going to:

- Will hop on just to QA things
- Will update the splash screen animation
- Do some minor code clean up

Teddarific (changes requested):
A few things we need to update:

1. Pull the sidebar into a shared layout — it might be worth migrating our NextJS projects to use the `app` directory for layouts: https://beta.nextjs.org/docs/routing/pages-and-layouts#layouts . Need to make sure that if the sidebar gets collapsed, it stays collapsed on page changes

2. I noticed switching tabs is slightly delayed on web — is this because its trying to fetch from localstorage? Not in scope of this PR, but wondering how we could relieve that issue.

This is looking awesome Karanvir, thanks for working with all our feedback, definitely a hugeeee feature so excited for this one. You are setting the foundation for Mirrorful 2.0!

karanvirsb:
> A few things we need to update:
> 
> 1. Pull the sidebar into a shared layout — it might be worth migrating our NextJS projects to use the `app` directory for layouts: https://beta.nextjs.org/docs/routing/pages-and-layouts#layouts . Need to make sure that if the sidebar gets collapsed, it stays collapsed on page changes
> 2. I noticed switching tabs is slightly delayed on web — is this because its trying to fetch from localstorage? Not in scope of this PR, but wondering how we could relieve that issue.
> 
> This is looking awesome Karanvir, thanks for working with all our feedback, definitely a hugeeee feature so excited for this one. You are setting the foundation for Mirrorful 2.0!

Hey Teddy I have fixed the sidebar issue and moved the LayoutWrapper into the layouts. So it's a shared component that does not rerender so the state stays the same. Also, I believe that the switch tab happens only in the dev environment and only on the first page load. I ran the build and started the server. In the actual build, there is no lag.

Teddarific (approved):

PR #266, opened 2023-03-30
Add copy to clipboard function for shadows

joelcmk:
Screen Shot 2023-03-30 at 3 57 11 PM


alexdanilowicz:
@joelcmk fyi I made a change to remove the jumpiness. Shown here: https://www.loom.com/share/2443386335c245febe4d9afbb8683d42

I move the padding to parent, instead of child.

Issue #195, opened 2023-03-20
Add fontWeight token 


alexdanilowicz:
Along with fontSize, fontWeight should be a token in the Typography section.

This issue is very similar to https://github.com/Mirrorful/mirrorful/issues/194

DivyanshuelRed:
Hey @alexdanilowicz, I'm Divyanshu, Can you assign this one to me?

alexdanilowicz:
Sure, go for it @thealoneshadow!

alexdanilowicz:
Hey @Bismay5467, yes! Would you like to push up a PR?

PR #277, opened 2023-04-01
fontWeight Token added in Typography Section

Bismay5467:
**Feature implemented : #195** 
Hey @alexdanilowicz ! Can you please review this pull request ?
The intention of this pull request is to add the `fontWeight` token in the Typography section.

### DIFF
--- a/packages/core/src/components/Typography/FontSize/EditFontSizeModal.tsx
+++ b/packages/core/src/components/Typography/FontSize/EditFontSizeModal.tsx
@@ -1,0 +1,165 @@
+import {
+  Box,
+  Button,
+  FormControl,
+  FormLabel,
+  Input,
+  Modal,
+  ModalBody,
+  ModalCloseButton,
+  ModalContent,
+  ModalFooter,
+  ModalHeader,
+  ModalOverlay,
+  Select,
+  Text,
+} from '@chakra-ui/react'
+import { useDisclosure } from '@chakra-ui/react'
+import { AlertDialogDelete } from '@core/components/AlertDialogDelete'
+import { TFontSizeVariant } from '@core/types'
+import { useEffect, useState } from 'react'
+
+import { fontUnits } from '../TypographyConstants'
+
+export function EditFontSizeModal({
+  isOpen,
+  isAdding,
+  onClose,
+  initialFontSizeVariant,
+  onUpdateFontSizeVariant,
+  onDeleteFontSizeVariant,
+}: {
+  isOpen: boolean
+  isAdding: boolean
+  onClose: () => void
+  initialFontSizeVariant?: TFontSizeVariant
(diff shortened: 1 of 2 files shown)


Teddarific:
Hey @Bismay5467! 

Awesome stuff here, just some high level comments:

1. We should probably let users define their own custom names for font weights, e.g. being able to rename "hairline" to whatever they want
2. I don't think Font Weights should be tied to Font Sizes, i.e. Font Weights should be a separate object to let people mix-and-match!

Teddarific:
@Bismay5467  I think a new section underneath `Font Sizes` would be ideal! Something like this:

image

The motivation behind this is that we want to provide a place for people to simply define values that they can choose how to mix-and-match them. We have plans for a theme editor where people can then use these to define bigger themes if that makes sense!

alexdanilowicz:
@Teddarific — thanks for chiming in. @Bismay5467 apologies that when were discussing in the issue, I didn't mention that font weight should be its own object (and NOT be tied in with fontsizes). That's on me and I apologize for the confusion.

I think what you have is a great start,  just need to make it its own object/section. Let us know if @Teddarific helped clarify it.

alexdanilowicz:
@Bismay5467 I'm a bit confused by the merge conflicts here? Why did the docs change and looks like the dashboard file was modified in main. Are you able to clean up the merge conflicts? Thanks!

docs/home/export-formats/colors.mdx
```diff
@@ -5,46 +5,46 @@ description: 'After configuring your theme in the editor, you can export colors
...
 
 ```javascript
 <button style={{ backgroundColor: 'var(--color-primary)'}}>
 ```
 
-## Using SCSS
-
-```scss
+````scss SCSS
+```scss SCSS
```
alexdanilowicz, line 22: None of the docs should change

packages/core/src/components/Dashboard/index.tsx
```diff
@@ -76,7 +78,7 @@ export function Dashboard({
         }
 
         setColors(data.tokens.colorData ?? [])
-        setTypography(data.tokens.typography)
```
alexdanilowicz, line 79: In main, the Dashboard index file was refactored. so this needs to be updated with what's on main

packages/core/src/components/Onboarding/pages/ExportSettings.tsx
```diff
@@ -35,7 +35,7 @@ export function ExportSettings({
   const shades = generateDefaultColorShades(primaryColor)
 
   return (
-    <Box css={{ display: 'flex', height: '100%' }}>
+    <Box css={{ display: 'flex', height: '100%' }} as="form">
```
alexdanilowicz, line 38: When main is merged, this diff should not show anymore

PR #279, opened 2023-04-02
adjust current shadow input UI

### LINKED ISSUE #257: Shadow sliders and shadow color picker won't update when modifying input value
When the user updates the variant value input, it should update the color picker/siders as well. Here is a loom of the issue. 

https://www.loom.com/share/3682b89eefa24a4faa59cc73cbf7a17e

ktra99:
resolves #257

# Notes

- Changed `onChangeEnd` to `onChange` I prefer to visually see the change that I make as I'm changing the values instead of having to wait until I drop the picker to view the change. Let me know if there's a thought process behind having `onChangeEnd`
- Wrapped `codeResult` into a `useMemo` hook to prevent unnecessary re-renders and recalculations
- Uplifting all state from `ShadowColorPicker` to `EditShadowModal` to allow both components to consume the necessary state variables
- Split the input field to multiple fields that handles their own individual value

# Sidenote
There's a lot of prop drilling in the current codebase. I know that there were some discussion around using zustand. I hope to start moving towards a state management solution as mirrorful grows :rocket:

### DIFF
--- a/packages/core/src/components/Shadows/EditShadowModal.tsx
+++ b/packages/core/src/components/Shadows/EditShadowModal.tsx
@@ -16,7 +16,8 @@ import {
 import { useDisclosure } from '@chakra-ui/react'
 import { AlertDialogDelete } from '@core/components/AlertDialogDelete'
 import { TShadowData } from '@core/types'
-import { useEffect, useState } from 'react'
+import { useEffect, useState, useMemo } from 'react'
+import { RgbColor } from '@hello-pangea/color-picker'
 
 import { ShadowColorPicker } from './ShadowColorPicker'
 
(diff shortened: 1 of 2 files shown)


packages/core/src/components/Shadows/EditShadowModal.tsx
```diff
@@ -78,6 +79,54 @@ export function EditShadowModal({
...
+  const [color, setColor] = useState(presetColor)
+  const [inputColor, setInputColor] = useState(presetColor)
+  const [hOffset, sethOffset] = useState(initialValues?.hOffset ?? 0)
+  const [vOffset, setVOffset] = useState(initialValues?.vOffset ?? 0)
+  const [blur, setBlur] = useState(initialValues?.blur ?? 0)
+  const [spread, setSpread] = useState(initialValues?.spread ?? 0)
+
+  const codeResult = useMemo(() => {
+    return `${hOffset}px ${vOffset}px ${blur}px ${spread}px ${color}`
+  }, [hOffset, vOffset, blur, spread, color])
```
Teddarific, line 91:
```suggestion
  const codeResult = `${hOffset}px ${vOffset}px ${blur}px ${spread}px ${color}`
```

nit: I think this is an instance where `useMemo`  might actually be less performant because of the extra memory / isn't actually saving too much computation

Teddarific (approved):
One small nit, but looks great to me, awesome job Kenny! 

I can see how the architecture is getting quite gnarly with multiple things controlling multiple states haha

@karanvirsb is working on adding Zustand, exciting stuff!

PR #282, opened 2023-04-04
improv: wrap onboarding steps in form elements and make them keyboard accessible

### LINKED ISSUE #267: Convert the onboarding steps into `<form />` elements
Let's convert each of the pages in `Onboarding` to a `<form />` with an `onSubmit` so hitting enter works.

Pranav2612000:
Previously, the onboarding was not keyboard accessible - Pressing enter did not move to the next step. This commit wraps the onboarding steps in a form component and changes the button to a submit so that the whole flow is `enter`-accessible. Also to prevent default form submit behaviour, we also need to add e.preventDefault() on the button clicks

Fixes #267

### DIFF
--- a/packages/core/src/components/Onboarding/pages/ExportSettings.tsx
+++ b/packages/core/src/components/Onboarding/pages/ExportSettings.tsx
@@ -35,7 +35,7 @@ export function ExportSettings({
   const shades = generateDefaultColorShades(primaryColor)
 
   return (
-    <Box css={{ display: 'flex', height: '100%' }}>
+    <Box css={{ display: 'flex', height: '100%' }} as="form">
       <Box
         width="50%"
         padding="12px"
@@ -80,11 +80,14 @@ export function ExportSettings({
             padding="8px 36px"
             size="lg"
             rightIcon={<ArrowForwardIcon />}
-            onClick={() => {
+            onClick={(e) => {
+              e.preventDefault()
               onExport()
               onUpdatePage(6)
             }}
             isDisabled={!fileTypes.length}
+            type="submit"
+            autoFocus
           >
             Next
           </Button>
--- a/packages/core/src/components/Onboarding/pages/ExportSettings.tsx
+++ b/packages/core/src/components/Onboarding/pages/ExportSettings.tsx
@@ -87,7 +87,6 @@ export function ExportSettings({
             }}
             isDisabled={!fileTypes.length}
             type="submit"
-            autoFocus
           >
             Next
           </Button>
--- a/packages/core/src/components/Onboarding/pages/ImportInstructions.tsx
+++ b/packages/core/src/components/Onboarding/pages/ImportInstructions.tsx
@@ -33,7 +33,7 @@ export function ImportInstructions({
   const shades = generateDefaultColorShades(primaryColor)
 
   return (
-    <Box css={{ display: 'flex', height: '100%', flexGrow: 1 }}>
+    <Box css={{ display: 'flex', height: '100%', flexGrow: 1 }} as="form">
       <Box
         css={{
           padding: '12px',
@@ -89,9 +89,12 @@ export function ImportInstructions({
             padding={'8px 36px'}
             size="lg"
             rightIcon={<ArrowForwardIcon />}
-            onClick={() => {
+            onClick={(e) => {
+              e.preventDefault()
               onUpdatePage(7)
             }}
+            type="submit"
+            autoFocus
           >
             Next
           </Button>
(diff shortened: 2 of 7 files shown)


Pranav2612000:
Some things worth discussing
1. Since we are using an external package ( @hello-pangea/color-picker ) for picking the primary color. It does not have first class accessiblity support. Choosing a color through the picker and pressing Enter has no action. But pressing enter from the input fields in the color picker does submit the form.

2. Clicking on paragraphs and other elements in the form and pressing enter has no action. I'm not sure if there is a HTML way of accomplishing this.

If we want these to be fixed we can do that by using javascript, useRefs and a React'ive way of doing things, but it would also add to the complexity and there are a lot of edge cases to be considered. Let me know what you folks think.

alexdanilowicz:
@Pranav2612000 we are 99% going to use a different color picker component, so no need to do anything regarding the Color Picker. I want to rewrite that component.

packages/core/src/components/Onboarding/pages/ExportSettings.tsx
```diff
@@ -80,11 +80,14 @@ export function ExportSettings({
...
             rightIcon={<ArrowForwardIcon />}
-            onClick={() => {
+            onClick={(e) => {
+              e.preventDefault()
               onExport()
               onUpdatePage(6)
             }}
             isDisabled={!fileTypes.length}
+            type="submit"
+            autoFocus
```
Teddarific, line 90:
```suggestion
```

packages/core/src/components/Onboarding/pages/ImportInstructions.tsx
```diff
@@ -89,9 +89,12 @@ export function ImportInstructions({
             padding={'8px 36px'}
             size="lg"
             rightIcon={<ArrowForwardIcon />}
-            onClick={() => {
+            onClick={(e) => {
+              e.preventDefault()
               onUpdatePage(7)
             }}
+            type="submit"
+            autoFocus
```
Teddarific, line 97:
```suggestion
```

packages/core/src/components/Onboarding/pages/NamePrimary.tsx
```diff
@@ -93,6 +94,8 @@ export function NamePrimary({
               onUpdatePrimaryName(name)
               onUpdatePage(3)
             }}
+            type="submit"
+            autoFocus
```
Teddarific, line 98:
```suggestion
```

packages/core/src/components/Onboarding/pages/OtherColors.tsx
```diff
@@ -107,6 +108,8 @@ export function OtherColors({
                 onUpdatePage(5)
               }
             }}
+            type="submit"
+            autoFocus
```
Teddarific, line 112:
```suggestion
```

packages/core/src/components/Onboarding/pages/PickPrimary.tsx
```diff
@@ -69,10 +69,13 @@ export function PickPrimary({
...
             size="lg"
             rightIcon={<ArrowForwardIcon />}
-            onClick={() => {
+            onClick={(e) => {
+              e.preventDefault()
               onUpdatePrimaryColor(primaryColor)
               onUpdatePage(2)
             }}
+            type="submit"
+            autoFocus
```
Teddarific, line 78:
```suggestion
```

packages/core/src/components/Onboarding/pages/Referral.tsx
```diff
@@ -126,6 +126,8 @@ export function Referral({
 
               onFinish()
             }}
+            type="submit"
+            autoFocus
```
Teddarific, line 130:
```suggestion
```

packages/core/src/components/Onboarding/pages/ReviewPrimary.tsx
```diff
@@ -94,9 +94,12 @@ export function ReviewPrimary({
             padding={'8px 36px'}
             size="lg"
             rightIcon={<ArrowForwardIcon />}
-            onClick={() => {
+            onClick={(e) => {
+              e.preventDefault()
               onUpdatePage(4)
             }}
+            type="submit"
+            autoFocus
```
Teddarific, line 102:
```suggestion
```

Teddarific (approved):
Awesome work @Pranav2612000 and appreciate the call outs!

Recognize that enter doesn't work on most of the pages because most of the pages aren't actually forms! In hindsight, I think the only screen that matters is the one where you have to type, aka the one with an input.

I removed the `autofocus` because don't think it's necessary! Awesome stuff, will merge in shortly. Welcome to the contributors list!

Pranav2612000:
> Awesome work @Pranav2612000 and appreciate the call outs!
> 
> Recognize that enter doesn't work on most of the pages because most of the pages aren't actually forms! In hindsight, I think the only screen that matters is the one where you have to type, aka the one with an input.
> 
> I removed the `autofocus` because don't think it's necessary! Awesome stuff, will merge in shortly. Welcome to the contributors list!

Thank you @Teddarific and @alexdanilowicz

PR #285, opened 2023-04-05
updated to take into account missing dist folder

karanvirsb:
This is related and a fix for this issue https://mirrorful.slack.com/archives/C04QKTVD8FP/p1680718339754139.

### DIFF
--- a/packages/mirrorful/package.json
+++ b/packages/mirrorful/package.json
@@ -18,7 +18,7 @@
     "test": "echo \"Error: no test specified\" && exit 1",
     "build": "tsc --build",
     "local": "run-script-os",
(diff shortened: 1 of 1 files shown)


alexdanilowicz:
I don’t think those commands are actually ever used. I think all the scripts should be deleted from that package.json because the only package.jsons that are used are the ones in the editor and in the core.

karanvirsb:
> I don’t think those commands are actually ever used. I think all the scripts should be deleted from that package.json because the only package.jsons that are used are the ones in the editor and in the core.

Okay as Teddy mentioned Ill delete the local scripts so they won't interfere.

alexdanilowicz:
updating the thread here, these are used for testing the cli locally. 

Thanks to @karanvirsb for adding a fix for windows users.

Issue #270, opened 2023-03-31
Give color a default name in the onboarding flow


Teddarific:
We can use https://github.com/colorjs/color-namer to pre-populate a suggested name in the second page of the onboarding flow.

alexdanilowicz:
Context: <img width="1052" alt="Screen Shot 2023-03-30 at 9 01 08 PM" src="https://user-images.githubusercontent.com/29822597/229020137-89e7aac2-ba3e-4d81-95fd-35a01bf8abc1.png">

iamoghenerukevwe:
I'll take this  up

prichter:
Could also potentially pick the nearest color name from the list we use in generating default colors, thus avoiding adding another library?

Pranav2612000:
Is anyone working on this? I'll like to pick this up.

JoshuaRotimi:
@alexdanilowicz @Teddarific I'm Trying to work on this and push a new branch for the feature. Seems like I need permission to create a PR

alexdanilowicz:
@JoshuaRotimi — hi, you need to create a fork of the repo. As is the case with most open-source repos, contributors must fork the repo.

In case this helps:
- https://reflectoring.io/github-fork-and-pull/#:~:text=Creating%20a%20fork%20on%20GitHub,to%20the%20fork%20on%20GitHub.
- https://docs.github.com/en/get-started/quickstart/fork-a-repo

The easiest thing to do here is to reclone the repo as a fork using the instructions above, and then copy paste over your changes.

Alternatively, you can convert your existing local copy to a fork:

- https://gist.github.com/jagregory/710671

JoshuaRotimi:
Ah, Thanks for pointing that out @alexdanilowicz. Just made the fix and created a pull request now.

alexdanilowicz:
>  Could also potentially pick the nearest color name from the list we use in generating default colors, thus avoiding adding another library?

This was spot on. After playing around with the library, I was pretty unimpresed with the names. For example, one of the names was "mediumpurple" which is very hard to split out. And if it resembled black, it would just be black. So, I ended using your awesome algorithm @prichter .

@JoshuaRotimi great work on this and thanks for helping start it! Same to you @Pranav2612000 for the review. 

I'm merging this in now. Will also post a write up in Slack.

alexdanilowicz:
Resolved by #288

PR #288, opened 2023-04-06
pre-populate onboarding primary color

JoshuaRotimi:
This PR fix issue 270 https://github.com/Mirrorful/mirrorful/issues/270.

A similar colour name is suggested based on the Onboarding Primary colour selected in Step 1.

Here's a short video showing how it works. https://vimeo.com/815414449/b02661cf55

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,5 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Editor-based HTTP Client requests
+/httpRequests/
--- a/packages/core/src/components/Onboarding/index.tsx
+++ b/packages/core/src/components/Onboarding/index.tsx
@@ -7,7 +7,7 @@ import {
   TConfig,
   TExportFileType,
 } from '@core/types'
-import { useState } from 'react'
+import { useEffect, useState } from 'react'
 
 import { TPlatform } from '../Layout'
 import { OnboardingContainer } from './OnboardingContainer'
@@ -19,6 +19,7 @@ import { PickPrimary } from './pages/PickPrimary'
 import { Referral } from './pages/Referral'
 import { ReviewPrimary } from './pages/ReviewPrimary'
 import { Welcome } from './pages/Welcome'
+var namer = require('color-namer')
 
 export function Onboarding({
   postStoreData,
@@ -35,6 +36,14 @@ export function Onboarding({
   const [fileTypes, setFileTypes] = useState<TExportFileType[]>(defaultFiles)
 
   const [page, setPage] = useState<number>(platform === 'web' ? 1 : 0)
+
+  let names = namer(primaryColor)
+
+  useEffect(() => {
+     names = namer(primaryColor)
+    setPrimaryName(names.html[0].name)
+
+  }, [primaryColor])
 
   const handleExport = async (
     primaryColorHex: string,
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,0 +1,63 @@
+<component name="ProjectCodeStyleConfiguration">
+  <code_scheme name="Project" version="173">
+    <HTMLCodeStyleSettings>
+      <option name="HTML_SPACE_INSIDE_EMPTY_TAG" value="true" />
+      <option name="HTML_QUOTE_STYLE" value="Single" />
+      <option name="HTML_ENFORCE_QUOTES" value="true" />
+    </HTMLCodeStyleSettings>
+    <JSCodeStyleSettings version="0">
+      <option name="USE_SEMICOLON_AFTER_STATEMENT" value="false" />
+      <option name="FORCE_SEMICOLON_STYLE" value="true" />
+      <option name="SPACE_BEFORE_FUNCTION_LEFT_PARENTH" value="false" />
+      <option name="USE_DOUBLE_QUOTES" value="false" />
+      <option name="FORCE_QUOTE_STYlE" value="true" />
+      <option name="ENFORCE_TRAILING_COMMA" value="WhenMultiline" />
+      <option name="SPACES_WITHIN_OBJECT_LITERAL_BRACES" value="true" />
+      <option name="SPACES_WITHIN_IMPORTS" value="true" />
+    </JSCodeStyleSettings>
+    <TypeScriptCodeStyleSettings version="0">
+      <option name="USE_SEMICOLON_AFTER_STATEMENT" value="false" />
+      <option name="FORCE_SEMICOLON_STYLE" value="true" />
+      <option name="SPACE_BEFORE_FUNCTION_LEFT_PARENTH" value="false" />
+      <option name="USE_DOUBLE_QUOTES" value="false" />
+      <option name="FORCE_QUOTE_STYlE" value="true" />
+      <option name="ENFORCE_TRAILING_COMMA" value="WhenMultiline" />
+      <option name="SPACES_WITHIN_OBJECT_LITERAL_BRACES" value="true" />
+      <option name="SPACES_WITHIN_IMPORTS" value="true" />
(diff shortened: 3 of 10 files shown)


.idea/.gitignore
Pranav2612000, line 1: Maybe @alexdanilowicz can answer this better, but I don't think you need to push these .idea/ files which have your editor config to the project.
You can use git reset/rebase or git commit --amend to modify the last commit and remove these files from the PR

packages/core/src/components/Onboarding/index.tsx
```diff
@@ -36,6 +37,14 @@ export function Onboarding({
 
   const [page, setPage] = useState<number>(platform === 'web' ? 1 : 0)
 
+  let names = namer(primaryColor)
```
Pranav2612000, line 40: Do we need to define this variable here? I guess if its used only inside the useEffect, we can just define and use it there?
Can you please check once if it works this way?

packages/core/src/components/Onboarding/index.tsx
```diff
@@ -36,6 +37,14 @@ export function Onboarding({
...
+
+  useEffect(() => {
+     names = namer(primaryColor)
+    setPrimaryName(names.html[0].name)
```
Pranav2612000, line 44: I know that the library guarantees that `names.html[0].name` would always exist, but maybe we could use optional chaining to ensure that this value exists and the page doesn't error out if some of these values are undefined.
Again, I don't think it's necessary but it's just a nit from my Typescript love. What are your thoughts @alexdanilowicz ?

packages/core/src/components/Onboarding/index.tsx
```diff
@@ -36,6 +37,14 @@ export function Onboarding({
 
   const [page, setPage] = useState<number>(platform === 'web' ? 1 : 0)
 
+  let names = namer(primaryColor)
+
+  useEffect(() => {
+     names = namer(primaryColor)
```
Pranav2612000, line 43: Nit: I think there's an extra space before this line. ` names = ....`

packages/core/src/components/Onboarding/index.tsx
```diff
@@ -19,6 +19,7 @@ import { PickPrimary } from './pages/PickPrimary'
 import { Referral } from './pages/Referral'
 import { ReviewPrimary } from './pages/ReviewPrimary'
 import { Welcome } from './pages/Welcome'
+var namer = require('color-namer')
```
Pranav2612000, line 22: Does this library not work with module imports ( the other imports we have - `import X from Y`) instead of using the commonjs ( `var x = require('y')` ) ones? In any case I also think it should be `const namer = ` instead of `var namer = ` ( The official documentation says var, but its 4 years old now. `const` is definitely better here )

Pranav2612000:
Hey @JoshuaRotimi I just took a look at the video - it looks amazing 😍  Can't wait for this to get merged ( It's a good thing I didn't pick it up - I wouldn't have been able to do it so quickly )
I've left some minor comments on the PR. Maybe you can address them if you feel they're appropriate and @alexdanilowicz agrees.

.idea/.gitignore
alexdanilowicz: Yeah let's just delete the .idea folder

packages/core/src/components/Onboarding/index.tsx
alexdanilowicz, line 22: Yeah can we use es6 import syntax here?

packages/core/src/components/Onboarding/index.tsx
alexdanilowicz: Agreed — i dont think this needs to be defined here.

packages/core/src/components/Onboarding/index.tsx
alexdanilowicz: Yes, I also think names should be defined here.

packages/core/src/components/Onboarding/index.tsx
alexdanilowicz: yeah, I think we should do somechaining and just set it to a blank string worse case. or wrap in a try catch

alexdanilowicz (commented):
Great job Joshua!!!

packages/core/src/components/Onboarding/index.tsx
JoshuaRotimi: Thanks guys @alexdanilowicz @Pranav2612000. I will make those changes and push again.

packages/core/src/components/Onboarding/index.tsx
```diff
@@ -36,6 +37,11 @@ export function Onboarding({
 
   const [page, setPage] = useState<number>(platform === 'web' ? 1 : 0)
 
+  useEffect(() => {
```
Teddarific, line 40: A `useEffect` is not ideal here — it's generally suspicious using a `useEffect` to achieve a "dependency effect", e.g. like achieving a "if X happens, then do Y". This can have some weird side effects!

Instead, we can put this logic directly on line 83, where the `primaryColor` gets set, so whenever the `primaryColor` gets set, we also just set the name there! Let me know if that makes sense — awesome job!

Teddarific (changes requested):
Hey @JoshuaRotimi, thanks for the PR! This has been a highly requested feature!

Left a comment about where this should be implement!

packages/core/src/components/Onboarding/index.tsx
```diff
@@ -19,6 +19,7 @@ import { PickPrimary } from './pages/PickPrimary'
 import { Referral } from './pages/Referral'
 import { ReviewPrimary } from './pages/ReviewPrimary'
 import { Welcome } from './pages/Welcome'
+const namer = require('color-namer')
```
Teddarific, line 22: Btw, looks like the ES6 `import` change didn't go through for some reason!

packages/core/src/components/Onboarding/index.tsx
JoshuaRotimi: Ah, that makes sense. I will implement that now.

packages/core/src/components/Onboarding/index.tsx
JoshuaRotimi: Oh, Could be because the npm library does not provide any typings. I found a solution which I will deploy in the next commit. Let me know if that works.

JoshuaRotimi:
The deployment and workflow build keep failing. I think I may be missing something. Any ideas? @alexdanilowicz

Pranav2612000:
Were you able to build it locally? I think it's possible the build is failing because we aren't using the ES6 import of the library correctly.
That being said, I looked at the library source code and it isn't a lot, just one distance calculating function from 'chromajs' and a list of colours. What do you think about building the same logic in our project instead of using the dependency?
Pro's:
- No need to unnecessary download other color types 
- Control over color names, number of color names, delta function etc
- Less dependency to worry about

Con:
- We need to maintain this function, we can't just assume it'll always work.

I think the single con is as strong as the multiple pro's. But if you think we want it in-house. I'll be happy to make it ready this weekend.

packages/core/src/components/Onboarding/index.tsx
```diff
@@ -36,6 +37,12 @@ export function Onboarding({
 
   const [page, setPage] = useState<number>(platform === 'web' ? 1 : 0)
 
+  const updatePrimaryColor = (newColor: string) => {
+    setPrimaryColor(newColor)
+    const names = namer(newColor)
+    setPrimaryName(names?.html[0]?.name)
```
Pranav2612000, line 43: Haven't tested it out, but will my previous name be overwritten by a default one if I change a color?  And if it does, is that expected?

JoshuaRotimi: Yes the user can change the default color suggested if they want to. They should be able to choose their preferred color name.

Issue #305, opened 2023-04-11
New variant should save on keyboard press enter


alexdanilowicz:
<img width="634" alt="Screen Shot 2023-04-11 at 12 35 25 PM" src="https://user-images.githubusercontent.com/29822597/231270086-0f6034ed-7aa1-4fe2-81aa-261e0cb452d9.png">

nagy-nabil:
downloading the deps rn and will give it a try

PR #309, opened 2023-04-12
save new variant with enter (#305)

nagy-nabil:
Issue #305 
added event listener to the window when the model is open on `keydown` event which trigger the save functionality if the key is `Enter`


alexdanilowicz:
Screen Shot 2023-04-11 at 11 24 24 PM

Hm, this isn't actually working for me. 

I also feel like the solution is instead to use `onKeyPress` with the `<Input>` component instead of event listeners. Want to give that a shot @nagy-nabil ?

Thanks for the PR! Glad to have you here.

 https://stackoverflow.com/questions/27827234/how-to-handle-the-onkeypress-event-in-reactjs

nagy-nabil:
i over thinked it and wanted the event to be fired when  the modal is open in general not only when there's input in focus.
sorry I give it another try with the same idea but this time, I handle the event on `ModalContent` component with tabIndex attribute, so please give it a try
but you're probably right and the event need to handled in the input component

```
Keyboard events are only generated by <input>, <textarea>, <summary> and anything with the contentEditable or tabindex attribute. If not caught, they [bubble](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Events#event_bubbling) up the [DOM tree](https://developer.mozilla.org/en-US/docs/Web/API/Document_object_model/Using_the_Document_Object_Model#what_is_a_dom_tree) until they reach [Document](https://developer.mozilla.org/en-US/docs/Web/API/Document).
```

Teddarific:
Hey @nagy-nabil ! Quick tip: the easiest way to do this is probably just wrap all the inputs with a `<form onSubmit={() => {...}}> .... </form>` 

Now, whenever inside the form, if the user hits enter, it'll call the `onSubmit` function on the form. Learning this changed my life haha

nagy-nabil:
yeah sure I could give it a try.

and just for learning purposes did you check the method I tried in last commit @alexdanilowicz  @Teddarific

alexdanilowicz:
@nagy-nabil — yes, I think it makes the most sense to do a `<Form onSubmit={() => {...}}> .... </Form>` and then change the button to have `type={"submit"}`. 

Your last commit adds the listener to the ModalContent, which I think is a bit anti-pattern.

nagy-nabil:
i wrapped all `FormControl` around one form and it did work with one expectation 
it doesn't recognize the `ColorPicker` as form element and from MDN the enter by default works only while editing form elements
```
The submit event fires when the user clicks a [submit button](https://developer.mozilla.org/en-US/docs/Glossary/Submit_button) or presses Enter while editing a field (e.g. [<input type="text">](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text)) in a form. The event is not sent to the form when calling the [form.submit()](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit) method directly.
```

I guess you should choose what is better for the project

and I can revert the non necessary commits and only leave one, Thank you

alexdanilowicz:
Interesting, thanks for sharing the bit about MDN only recognizing when form elements are in focus. I think what you have is great. Going to merge as-is, really great work! @nagy-nabil ! Thank you

nagy-nabil:
@alexdanilowicz  thank you, my first ever open source contribution btw thank you again!

PR #315, opened 2023-04-12
cannot have two variants with the same name (#306)

### LINKED ISSUE #306: Warn user that you cannot have variants of the same name
As a user, if I accidentally name a variant with a name that I already have, then it will overwrite the old one.

nagy-nabil:
Fix #306 
- before saving a new variant check if this name already exists to not overwrite it, if exists throw an error
- catch any errors in `EditVariantModal` that happens with the save handler and set the already existing error state to the thrown error message 

here's an example
image

### DIFF
--- a/packages/core/src/components/ColorPalette/ColorDisplay.tsx
+++ b/packages/core/src/components/ColorPalette/ColorDisplay.tsx
@@ -333,6 +333,10 @@ export function ColorDisplay({
         isOpen={isAddVariantModalOpen}
         onClose={onAddVariantModalClose}
         onUpdateVariant={(newVariant: TNamedToken) => {
+          // check if the new Variant name already exist, if exist throw error
+          if (newVariant.name in colorData) {
+            throw new Error('Variant Name Already Exist')
+          }
           const updatedColorData = { ...colorData }
           updatedColorData[newVariant.name] = newVariant.token
 
--- a/packages/core/src/components/ColorPalette/ColorDisplay.tsx
+++ b/packages/core/src/components/ColorPalette/ColorDisplay.tsx
@@ -335,7 +335,7 @@ export function ColorDisplay({
         onUpdateVariant={(newVariant: TNamedToken) => {
           // check if the new Variant name already exist, if exist throw error
           if (newVariant.name in colorData) {
-            throw new Error('Variant Name Already Exist')
+            throw new Error('This name already exists, please choose a different name.')
           }
           const updatedColorData = { ...colorData }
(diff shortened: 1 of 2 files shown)


packages/core/src/components/ColorPalette/ColorDisplay.tsx
```diff
@@ -333,6 +333,10 @@ export function ColorDisplay({
         isOpen={isAddVariantModalOpen}
         onClose={onAddVariantModalClose}
         onUpdateVariant={(newVariant: TNamedToken) => {
+          // check if the new Variant name already exist, if exist throw error
+          if (newVariant.name in colorData) {
+            throw new Error('Variant Name Already Exist')
```
alexdanilowicz, line 338:
```suggestion
            throw new Error('This name already exists, please choose a different name.')
```

packages/core/src/components/ColorPalette/EditVariantModal.tsx
```diff
@@ -74,7 +74,15 @@ export function EditVariantModal({
...
     setError(null)
-    onUpdateVariant(variant)
+    // catch any thrown errors on save
+    try {
+      onUpdateVariant(variant)
+    } catch (err) {
+      if (err instanceof Error) {
+        setError(err.message)
+        return
+      }
```
alexdanilowicz, line 84: instead of this and throwing an sett, can you just do 

```javascript
 if (newVariant.name in colorData) {
         setError('This name already exists, please choose a different name.')
 }
 ```

nagy-nabil: i don't think you have access to `colorData` in `EditVariantModal.tsx` so you cannot check before submitting 

also in `ColorDisplay.tsx`  don't have access to the error state because its initialization is in `EditVariantModal.tsx`  so you cannot change it in the parent

nagy-nabil: but i can see why you don't want to throw an error, I guess I can move the `error` state from the child to the parent component and pass it as prop to the child
- move `error` state to `ColorDisplay` to be controlled by the parent
- `EditVariantModal` -the child here- would take another prop which is the error state

alexdanilowicz: Ok, I guess it's fine as is.

alexdanilowicz: Thanks for your help @nagy-nabil! 🎊

Issue #250, opened 2023-03-29
Clean up box shadow editing UI to support complex shadows


alexdanilowicz:
As a user, if I have a box shadow  with multiple values (separated by commas), the editor UI breaks.

This is the result of https://github.com/Mirrorful/mirrorful/pull/248

If the box shadows value can be an array… then the editor needs a way to specify WHICH shadow you are editing. As the picker and sliders are only one array at a time. This is why Figma’s editor works because you only select one shadow at a time.

ktra99:
I would like to work on this issue together with #257 as I believe both of these issues are interconnected

joelcmk:
Hey Alex, I would like to help with this issue!

alexdanilowicz:
@joelcmk awesome! You are the expert on this. Sounds good! Thanks Joel.

joelcmk:
Awesome, before working on this! I would like to see if this is a good idea!

Now, the modal to edit/add a shadow looks like this. I don't think that it makes sense to have an input to edit the values when we already have sliders that do the same thing. Instead, we could add a div to show the output. 

<img width="533" alt="Screen Shot 2023-04-05 at 2 45 57 PM" src="https://user-images.githubusercontent.com/57118300/230189252-59687877-21fc-4a03-8557-a9faf90b1f0c.png">

I made a quick Figma to show how the changes would look like. I removed the inputs, add a button to add a new shadow (this would help for complex shadows), and added a box to input the result. 

<img width="485" alt="Screen Shot 2023-04-05 at 2 42 57 PM" src="https://user-images.githubusercontent.com/57118300/230191780-b2c18a98-c515-405c-81c7-bdc47bc1ca25.png">

Any feedback is appreciated!

Teddarific:
Hey @joelcmk! Thanks for the mock up, helps a ton discussing this feature.

I know it makes this feature itself much more complicated, but I think we still need to include a way to enter in a raw value. People will find examples or come with their own box shadows, in which case they will want to simply copy and paste if that makes sense!

I think I would prefer inputs over sliders as well actually, as sliders aren't great for precise measurements. Let me know what you think!

alexdanilowicz:
I agree with @Teddarific. Let us know if you have any thoughts here @joelcmk. Love how you originally implemented this, are now doing mockups, and then will take your feature to the next level! The true cycle of product development. Awesome stuff.

alexdanilowicz:
[...]

PR #325, opened 2023-04-14
Add support for complex shadows

joelcmk:
Fixes issue #250


packages/core/src/components/Shadows/ShadowsSection.tsx
```diff
@@ -177,6 +211,8 @@ export function ShadowsSection({
     onClose: onAddVariantModalClose,
   } = useDisclosure()
 
+  console.log(shadows)
```
alexdanilowicz, line 214:
```suggestion
```

alexdanilowicz:
@joelcmk — super nice work on this! Really good.

1. Wanna fix up the merge conflicts by pulling main?
2. @ktra99 FYI on this PR
3. I think a space is missing after adding a shadow (should be `, ` aka comma space) 
4. There's an edge case if you add too many shadows: 

Screen Shot 2023-04-15 at 6 16 31 PM

5. looks like there's a bug where you can't actually edit the value input box. not sure if that's intentional but i feel like expected behavior is being able to copy paste in anything into the input or 
 edit any part of it: inhttps://www.loom.com/share/7d52a94c58bf4b3f8135279b0637d733

joelcmk:
Hey Alex, thanks for the feedback. I'll work on those issues, just have a question about number 4. Should there be a limit for the number of shadows allowed?

alexdanilowicz:
> Should there be a limit for the number of shadows allowed?

Ideally, no, as I don't think there's an actual limit with CSS, but honestly i'd be shocked if we get any users with more than 4.  so id do whatever is easiest and leads to the least code bloat: limit of 4? 

If you want to be fancy, you could have arrows that show you 4 at a time and then it's infinite (no limit) Also, I think maybe you should number them if possible? Like the icon for 1 has a 1 near it or on top of it, 2, 3, 4....

^ that can be a fast follow though. I'd rather have an actual user say "i need more than 4 shadows".

Main thing right now is just making sure it doesn't look/appear buggy.

alexdanilowicz:
@joelcmk great work.

1. We might wanna do a regex to validate it's a valid box shadow? for example, you can input `a`. 

Screen Shot 2023-04-19 at 9 20 09 AM


2. so NICE work booking up Slider -> Input Box. ideally the input box to the slider values are also hooked up, otherwise it's kinda weird and feels buggy. Like if i change input box value to be "15px"for blur, it should change blur slider to 15px.

 
Screen Shot 2023-04-19 at 9 21 29 AM

3. I think the input box should only show one at a time. because the concept of "selecting a shadow" at the top suggests to me I'm only editing one shadow at a time. I think this will make 1 and 2 easier if you know the input box only has one shadow in it at a given time... (ill be honest, I'm not 100% sure because i would like user feedback to see if the user wants an input box to input all shadows at once.... but right now it feels like selecting one shadow at a time is confusing when the input box shows all of them.) 

Screen Shot 2023-04-19 at 9 23 35 AM

alexdanilowicz:
Hi, this is looking better and better each time i see it:

1. I think ideally it doesn't tell you its an invalid format until you save. Its kinda of weird to type normally and it yells at you its invalid. Think about how password checkers work. They usually don't tell you its too short or too long until after you submit. 
Screen Shot 2023-05-01 at 4 19 52 PM

2. If it's invalid, it shouldn't also say "please fill out all the fields". Also, seems like both error reds are a different color! 
Screen Shot 2023-05-01 at 4 22 02 PM

3. on that note, "Please fill out all fields".. it should highlight the name form red if that's missing, otherwise its confusing whats missing. like, what's the error? it's unclear.
Screen Shot 2023-05-01 at 4 27 58 PM


3. let's make the save match the color (yours is blue, should be grey) and placement (save to the right, delete to the left) of the color modal
Screen Shot 2023-05-01 at 4 24 03 PM
Screen Shot 2023-05-01 at 4 23 35 PM

4. it's still kind of confusing to me what these are. I think they should be labeled: 1, 2, 3, 4... ask your partner or friend if it makes sense to them what it is. 
Screen Shot 2023-05-01 at 4 25 18 PM

5. Instead of variant name, it should say shadow name. Also, it should match the size of the header in the color modal. "Variant name" is much larger in the color modal than your modal.

great work! it's so close. and these are simple things.

PR #327, opened 2023-04-15
Creating Standard UI Components

karanvirsb:


Teddarific (approved):
This is looking great Karanvir!! Had a few broad strokes, but it's tough because we never gave you a Figma or anything so you've been tossed in a world of ambiguity haha

I'll go ahead and merge this in a day or two, and we can continue hacking on this together! Thanks for writing all these storybook stories and stuff, it's going to be a huge help!

PR #340, opened 2023-04-20
fix: use fern registry for client

dsinghvi:
Messing with ts files is big brain so I'm moving Mirrorful over to consuming the client that is published to `npm.buildwithfern.com`. 

When you run `fern generate` the server will update. 

When you run `fern generate --group client` a new client will be published. The mirrorful package just brings in this client via a dependency in package.json.


packages/mirrorful/package.json
```diff
@@ -44,6 +44,7 @@
     "@emotion/react": "^11.10.6",
     "@emotion/styled": "^11.10.6",
     "@mirrorful/core": "2.0.5",
+    "@mirrorful-fern/api-client": "0.0.223",
```
dsinghvi, line 47: just bring in the api client as a dependency

fern/api/generators.yml
```diff
@@ -1,17 +1,20 @@
 default-group: local
 groups:
+  client:
```
dsinghvi, line 3: separate group for client generation `fern generate --group client`

dsinghvi:
@alexdanilowicz @Teddarific

PR #341, opened 2023-04-20
fix: move generated client into mirrorful package

dsinghvi:
Generates the mirrorful client in `packages/mirrorful/cli/client` and adds relevant dependencies to `package.json`


.gitattributes
```diff
@@ -1,3 +1,3 @@
 packages/server/src/api/generated/** linguist-generated linguist-vendored
-packages/client/** linguist-generated linguist-vendored
+packages/packages/mirrorful/cli/client/** linguist-generated linguist-vendored
```
alexdanilowicz, line 2:
```suggestion
packages/mirrorful/cli/client/** linguist-generated linguist-vendored
```

i think

dsinghvi: yes thankyou

alexdanilowicz:
My going to close this for now and potentially reopen if we end up wanting/needing a generated client version. Thanks for opening. For now, using the packaged version.

PR #342, opened 2023-04-20
run fern generate via docker in CI

dsinghvi:
Running via docker in CI will make the check a little faster.


alexdanilowicz:
Not sure why docker gives a different output than local. For now, I'll just close this PR but thanks for opening and good to know if feel need to speed it up down the road.

PR #343, opened 2023-04-20
feat: allow user to use custom placeholder to test out the typographies

### LINKED ISSUE #294: [FR] Allow user to try out Typographies on custom placeholder text
Right now, the user can just see the different Typography styles on the default `Lorem Ipsum` placeholder text. 
image

It would be helpful if the user is able to customize the placeholder text and actually be able to see how his text looks in the chosen font. ( Similar to google fonts ) Thoughts? @alexdanilowicz @Teddarific 

This could be implemented by adding an input box at the top of the page ( between Typography heading and Font-Sizes ) which allows user to customize the placeholder text.

Pranav2612000:
Fixes #294 

image

### DIFF
--- a/packages/core/src/components/Typography/FontSize/FontSizesSection.tsx
+++ b/packages/core/src/components/Typography/FontSize/FontSizesSection.tsx
@@ -5,9 +5,11 @@ import { DisplayFontProperty } from '../DisplayFontProperty'
 
 export function FontSizesSection({
   fontSizeData,
+  placeholder,
   onUpdateFontPropertyData,
 }: {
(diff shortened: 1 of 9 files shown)


packages/core/src/components/Typography/FontSize/FontSizesSection.tsx
Pranav2612000, line 1: Using prop drilling as suggested by @alexdanilowicz here https://github.com/Mirrorful/mirrorful/issues/294#issuecomment-1514071750

alexdanilowicz:
Looks great to me, merging!

PR #366, opened 2023-04-26
chore: add paths property to github action workflows

### LINKED ISSUE #355: Add `path` to workflows to ensure workflows only run when there are changes in their directories
For example, the server workflow should only run if the PR changes content in the packages/server directory.

Pranav2612000:
This will ensure that only the workflows relevant to the changes will be triggered, instead of all workflows running.

Ref:
- https://stackoverflow.com/questions/63822219/how-to-run-github-actions-workflow-only-if-the-pushed-files-are-in-a-specific-fo
- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths

Fixes #355

### DIFF
--- a/.github/workflows/web.yml
+++ b/.github/workflows/web.yml
@@ -5,6 +5,7 @@ on:
       - main
     paths:
       - 'apps/web/**'
+      - 'packages/core/**'
   push:
     branches:
       - main
--- a/.github/workflows/web.yml
+++ b/.github/workflows/web.yml
@@ -3,9 +3,13 @@ on:
   pull_request:
     branches:
       - main
+    paths:
+      - 'apps/web/**'
   push:
     branches:
       - main
+    paths:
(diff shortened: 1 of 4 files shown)


Pranav2612000:
I'm not exactly sure how we'll be able to test this out without merging. If anyone has any ideas will love to try them out.

Teddarific:
Hey @Pranav2612000! Testing this is definitely annoying, so honestly we can just merge it in and we'll notice if it works or not. Low stakes!

.github/workflows/web.yml
```diff
@@ -3,9 +3,13 @@ on:
   pull_request:
     branches:
       - main
+    paths:
+      - 'apps/web/**'
```
Teddarific, line 7:
```suggestion
      - 'apps/web/**'
      - 'packages/core/**'
```

Issue #364, opened 2023-04-25
If you rename a color, there is a jump (color moves to the bottom).


alexdanilowicz:
Colors should state in place. Or if there is some intentional sorting, it should sort on mount only.

PR #371, opened 2023-04-26
Retain the order of colors when renaming

tvergho:
Should fix [#364](https://github.com/Mirrorful/mirrorful/issues/364) by retaining the order of colors in the `colors` object.


alexdanilowicz:
Great work!

PR #380, opened 2023-04-28
Dark mode feature added to mirrorful editor

### LINKED ISSUE #61: Add dark mode to editor
My concussion turned me into a pseudo-vampire, so this is partly an a11y issue.

Please add darkmode support to both the designer, then dogfood it in your homepage.
TIA

Bismay5467:
Resolves #61


packages/core/src/components/AlertDialogDelete.tsx
```diff
@@ -36,7 +44,7 @@ export function AlertDialogDelete({
       onClose={onClose}
     >
       <AlertDialogOverlay>
-        <AlertDialogContent>
+        <AlertDialogContent bg={backgroundColor}>
```
Pranav2612000, line 47: I'm not sure if this is the right way to do it. This way we'll need to add a prop to every element we support and then drill this value to the element. Can we define the themes and the **current theme** in a top level context or zustand file somewhere and have that imported in the actual elements. This way we don't need to keep on passing the color. The actual component can refer this zustand store to identify what the currently chosen theme is and style it accordingly.

alexdanilowicz:
Hi, great first steps. Ideally, we use Mirrorful itself to power our dark mode themeing. @Pranav2612000 is correct in that it seems weird to pass props and instead should be top-level themeing, but even the top-level themeing is not perfect Let me talk to @Teddarific as we will be adding a semantic tab soon to Mirrorful and then can use Mirrorful in Mirrorful. As-is, this PR is a GREAT START (thanks @Bismay5467), but I don't think using zustand, nor manually defining the theme in each file is the perfect solution, so let's hold on off this for a bit. I'm going to update the issue.

alexdanilowicz:
Going to close this for the time being, in alignment with what was written above.

PR #387, opened 2023-05-02
Update .prettierignore to exclude examples dir

alexdanilowicz:


alexdanilowicz:
Discussion in: https://mirrorful.slack.com/archives/C04RR2D7C65/p1682985919532639?thread_ts=1678815500.691389&cid=C04RR2D7C65

PR #391, opened 2023-05-02
Add Link component to Blocks Component Library

alexdanilowicz:


packages/blocks/src/Link/index.tsx
```diff
@@ -1,34 +1,29 @@
...
+  isExternal?: boolean
+}) {
   return (
     <ChakraLink
-      css={{ display: 'flex', gap: 8, alignItems: 'center' }}
-      {...props}
+      as={NextLink}
+      isExternal={isExternal}
+      href={href}
+      color={'purple.500'}
```
Teddarific, line 23: Use CSS variables!

E.g. `var(--primary-color)`

Teddarific (approved):
nice, other than just using CSS variables, looks great!

PR #401, opened 2023-05-09
Fix/Typogarphy: Changed variant value type in edit modals from number to string

### LINKED ISSUE #394: Minor Bugs under Typography section
- If the placeholder value in the typography section is an empty string, the fontsize token will show the default content. Fontweight and lineheight tokens, on the other hand, do not display any default text.

- The default value for the fontsize and lineheights variation value input box in the edit or add modals is set to 0 and cannot be deleted when editing the input field. When entering a new value, the input is concatenated with the existing value i.e. 0, which can lead to a poor user experience. We can use a string type instead of a number type for variantValue. As in the fontweight edit(add) modal, we can simply display an error message for every non-numeric value entered. @alexdanilowicz Thoughts?

Bismay5467:
Resolves #394


packages/core/src/components/Typography/FontSize/EditFontSizeModal.tsx
```diff
@@ -60,15 +60,28 @@ export function EditFontSizeModal({
...
+  const variantValueRef = useRef() as MutableRefObject<HTMLInputElement>
 
   const handleSave = () => {
     setError(null)
-    if (variantName === '') {
+    if (variantName === '' || variantValue === '') {
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+      variantName === ''
+        ? variantNameRef.current.focus()
+        : variantValueRef.current.focus()
```
Pranav2612000, line 72:
```suggestion
    if (variantName !== '' && variantValue !== '') {
      return;
    } else if (variantName === '') {
      variantNameRef.current.focus();
    } else if (variantValue === '') {
      variantValueRef.current.focus()
    }
```

Pranav2612000: Personal opinion 😅 , but I feel it's easier to read this way. Feel free to resolve the comment if you don't agree.

Pranav2612000, line 72: Can we make this more readable? I think we can do the variantValue check in a different if condition.

alexdanilowicz:
@Bismay5467 — could you clarify, what is the intention of changing it from a string to a number?

alexdanilowicz:
Great work @Bismay5467!

PR #415, opened 2023-05-11
Users may now make inline edits to color names. 

### LINKED ISSUE #407: Colors Palette: Allow users to change color name by clicking on the Color name
image

For example, clicking on "Medium Purple" should show an input box that allows the user to modify the current name, instead of the user having to click on the pencil icon and then choose `Edit Color Name`

The user should be able to edit inline on the same page. Similar to how the UI of the HSL values boxes.

Bismay5467:
Resolves #407


packages/core/src/components/ColorPalette/ColorDisplay.tsx
```diff
@@ -124,7 +120,28 @@ export function ColorDisplay({
         mb={5}
         style={{ display: 'flex', flexDirection: 'row', alignItems: 'center' }}
       >
-        <Text style={{ fontWeight: 600, fontSize: '1.5rem' }}>{colorName}</Text>
+        <Input
```
Pranav2612000, line 123: I think the expected behaviour should be that the input box opens up only after double-clicking on the color name ( rather than single-click, which is the behaviour in this PR ). Thoughts @alexdanilowicz ?

packages/core/src/components/ColorPalette/ColorDisplay.tsx
```diff
@@ -133,9 +150,6 @@ export function ColorDisplay({
             <MenuItem onClick={() => onBaseModalOpen()}>
               Edit Base Color
             </MenuItem>
-            <MenuItem onClick={() => onColorNameModalOpen()}>
-              Edit Color Name
-            </MenuItem>
```
Pranav2612000, line 138: I think we should keep this way of updating name as well.

packages/core/src/components/ColorPalette/ColorDisplay.tsx
alexdanilowicz: Agreed @Pranav2612000.

packages/core/src/components/ColorPalette/ColorDisplay.tsx
alexdanilowicz: Agreed, yes let's keep the name

packages/mirrorful/editor/src/blocks.css
```diff
@@ -1,4 +1,4 @@
-@import '../.mirrorful/theme.css';
+/* @import '../.mirrorful/theme.css'; */
```
alexdanilowicz, line 1:
```suggestion
@import '../.mirrorful/theme.css';
```

@Bismay5467 why was this removed?

alexdanilowicz:
Building it without the removal of theme.scss in blocks so I can ensure it still works and build fine. then will merge this in.

Issue #416, opened 2023-05-11
Fontweight should be max of 900, not 1000.


alexdanilowicz:
1000 is invalid.

<img width="1321" alt="Screen Shot 2023-05-11 at 2 43 00 PM" src="https://github.com/Mirrorful/mirrorful/assets/29822597/ce222a8b-fe98-4bac-9b7e-7b9b8b058322">

PR #417, opened 2023-05-11
fix: reduced font weight allowed limits

asjadaugust:
Editing the font-weight above 900 or adding a similar weight would show error message. #416 
image


alexdanilowicz:
Hi @asjadaugust thank you so much for your PR! 🎊 

Welcome to the contributors list. I will merge this in right now.

PR #419, opened 2023-05-12
chore: update eslint config for web, core to show warnings for stray logs

Pranav2612000:
Reference: https://mirrorful.slack.com/archives/C04RR2D7C65/p1683770775020169?thread_ts=1683769489.423339&cid=C04RR2D7C65

This would prevent unnecessary console logs in our app. If we want to keep a specific log for debugging, we can always eslint-disable the corresponding line.


alexdanilowicz:
Nice, I'm cool with this. Great work @Pranav2612000

PR #420, opened 2023-05-12
Update Head and preload images

Teddarific:


apps/web/src/app/layout.tsx
```diff
@@ -108,6 +109,17 @@ export default function RootLayout({
...
+          <title>Mirrorful Editor</title>
+          <meta content="width=device-width, initial-scale=1" name="viewport" />
+          <meta
+            name="description"
+            content="Create, edit, and manage your app's theme."
+          />
+          <link rel="icon" href="/favicon.ico" />
+          <link rel="preload" href="/components_graphic_dark.png" />
+          <link rel="preload" href="/components_graphic_light.png" />
+        </Head>
```
alexdanilowicz, line 122: If this is here now, should we delete the Head from app/page.tsx? 

and also update the head in editor/index.tsx to this one, so they match on web and package?

PR #422, opened 2023-05-12
Fix onboarding

alexdanilowicz:


packages/core/src/components/ThemeProvider.tsx
```diff
@@ -10,6 +10,11 @@ const theme = extendTheme({
     },
   },
   components: {
+    Modal: {
+      dialog: {
+        bg: 'var(--background-color-primary)',
+      },
+    },
```
alexdanilowicz, line 17: this is not actually a problem. I thought it was a modal at first. should we just use charka defaults? @Teddarific

packages/core/src/components/ThemeProvider.tsx
```diff
@@ -34,5 +41,6 @@ export function MirrorfulThemeProvider({
 }: {
   children: React.ReactNode
 }) {
+  console.log('hi')
```
alexdanilowicz, line 44:
```suggestion
```

packages/core/src/components/ThemeProvider.tsx
Teddarific: Yeah, just I'd just delete this

Issue #408, opened 2023-05-10
Shadows: Change shadowpicker color picker to the new color picker 


Pranav2612000:
Update the Add/Edit Shadow Variant Modal to migrate away from the older color picker and use the newer color picker.

<img width="1013" alt="image" src="https://github.com/Mirrorful/mirrorful/assets/20909078/58a0d4b5-5b6d-4764-bcd1-2aa3047e61b1">

joelcmk:
I can help with this!

alexdanilowicz:
@joelcmk awesome, assigned to you! Thanks for picking this up! 🎊

PR #429, opened 2023-05-13
Updates shadows to new color picker

joelcmk:
Solves issue #408

### DIFF
--- a/packages/core/src/components/ColorPalette/ColorPicker.tsx
+++ b/packages/core/src/components/ColorPalette/ColorPicker.tsx
@@ -29,11 +29,13 @@ function DropdownInput({
   colorPickerColor,
   format,
   onSetFormat,
+  shadow,
 }: {
   colorPickerColor: AnyColor
   onChange: (color: AnyColor) => void
   format: TFormat
   onSetFormat: (format: TFormat) => void
+  shadow: boolean
 }) {
   const [hexInputValue, setHexInputValue] = useState<string>(
     tinycolor(colorPickerColor).toHexString()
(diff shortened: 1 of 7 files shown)


alexdanilowicz:
@joelcmk total nit, but fyi i think we have to say `closes #XX` (or "fixes" or "resolves") without the word `issue` for the PR to auto link to the issue.

packages/core/src/components/ColorPalette/ColorPicker.tsx
```diff
@@ -29,11 +29,13 @@ function DropdownInput({
   colorPickerColor,
   format,
   onSetFormat,
+  shadow,
```
alexdanilowicz, line 32:
```suggestion
  shadow = false,
```

packages/core/src/components/ColorPalette/ColorPicker.tsx
```diff
@@ -29,11 +29,13 @@ function DropdownInput({
...
 }: {
   colorPickerColor: AnyColor
   onChange: (color: AnyColor) => void
   format: TFormat
   onSetFormat: (format: TFormat) => void
+  shadow: boolean
```
alexdanilowicz, line 38:
```suggestion
  shadow?: boolean
```

packages/core/src/components/ColorPalette/ColorPicker.tsx
alexdanilowicz: this way we don't have to prop drill `shadows` everywhere. as 99% of cases don't need it

alexdanilowicz:
Nice job @joelcmk. Going to merge this in. I think as a fast follow we could add some padding. The % input (alpha) is a little tight, and in general the modal could use some "breathing room" (more whitespace) but I think this is mergeable. great work!
Screen Shot 2023-05-15 at 3 07 59 PM

PR #431, opened 2023-05-13
improv: make the onboarding flow mobile compatible

### LINKED ISSUE #409: Make the entire Mirrorful dashboard Mobile-friendly
We should leverage Charka's responsive styles support: https://chakra-ui.com/docs/styled-system/responsive-styles

Everything should be mobile friendly and responsive.

Pranav2612000:
Partially fixes #409

### DIFF
--- a/packages/core/src/components/Onboarding/OnboardingContainer.tsx
+++ b/packages/core/src/components/Onboarding/OnboardingContainer.tsx
@@ -47,9 +47,7 @@ export function OnboardingContainer({
       <Box
         css={{
           backgroundColor: 'var(--background-color-primary)',
-          padding: '64px 128px',
           borderRadius: 8,
-          height: '50%',
         }}
         shadow="2xl"
         width={{
@@ -57,15 +55,23 @@ export function OnboardingContainer({
           md: '80%',
           lg: '75%',
         }}
+        height={{
+          base: '100%',
+          md: '50%',
+        }}
         minHeight={'600px'}
-        minWidth={'700px'}
+        minWidth={{
+          md: '700px',
+        }}
         maxWidth={'1000px'}
         padding={{
           base: '24px',
-          md: '64px 64px',
+          md: '64px 128px',
           //   lg: '64px 128px',
         }}
-        maxHeight={'700px'}
+        maxHeight={{
+          md: '700px',
+        }}
       >
         {children}
       </Box>
--- a/packages/core/src/components/Onboarding/OnboardingContainer.tsx
+++ b/packages/core/src/components/Onboarding/OnboardingContainer.tsx
@@ -16,6 +16,10 @@ export function OnboardingContainer({
         justifyContent: 'center',
         alignItems: 'center',
       }}
+      minHeight={{
+        base: '896px',
+        md: 'unset',
+      }}
     >
       <Box
         css={{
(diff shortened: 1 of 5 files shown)


Pranav2612000:
@alexdanilowicz Can you help with reviewing this? I know there's scope for improvement, but this is just the first cut to get started with the PR.

packages/core/src/components/Onboarding/OnboardingContainer.tsx
```diff
@@ -39,33 +39,41 @@ export function OnboardingContainer({
...
           md: '80%',
           lg: '75%',
         }}
+        height={{
+          base: '100%',
+          md: '50%',
+        }}
         minHeight={'600px'}
-        minWidth={'700px'}
+        minWidth={[null, '700px']}
```
alexdanilowicz, line 65: so what's the idea here, `sm: is null` and then it's 700px for md and above? 

A nit but, I think I prefer the object syntax. Just makes it more readable: https://chakra-ui.com/docs/styled-system/responsive-styles#the-object-syntax

Just noting it for moving forward.

alexdanilowicz (commented):
Hey thanks for kicking this off! so excited for this! My main piece of feedback: I think for the onboarding flow, we should make the modal just cover the entire screen (or at least most of it). It's kinda of weird for the user to have to scroll (and to even be able to scroll that much). If this  is not easily possible though, no worries, I can take a look. 

Screen Shot 2023-05-14 at 9 25 27 PM

The other nit is just use object syntax.

alexdanilowicz:
Screen Shot 2023-05-14 at 9 25 27 PM

packages/core/src/components/Onboarding/OnboardingContainer.tsx
Pranav2612000: Done.

packages/core/src/components/Onboarding/pages/NamePrimary.tsx
```diff
@@ -101,12 +111,22 @@ export function NamePrimary({
         </Box>
       </Box>
       <Box
+        width={{
+          base: '100%',
+          md: '60%',
+        }}
+        order={{
+          base: -1,
+          md: 0,
```
Pranav2612000, line 120: Also wanted your thoughts on this @alexdanilowicz 
IMO, it looked better when the text was below the visual element.

Pranav2612000:
PS: The vercel "Comment" Floating Button is also causing some weird scroll issues. I would suggest to test with the button disabled for "this session", if possible.

packages/core/src/components/Onboarding/pages/NamePrimary.tsx
alexdanilowicz: I think it looks best in all instances when the text is ABOVE and then the content, and then the buttons.

alexdanilowicz: For ex, looks weird here. Let's just be consistent and put text above in all instances, and then this is ready to merge.

Screen Shot 2023-05-15 at 7 59 53 PM

Pranav2612000: Sure. I've done that change.

The button was part of the Text Container, so I had to `absolute` position it at the bottom of the page. I hope that's fine. 
( An alternate approach could be to make the Button an sibling of the Text and Content Container and change the order depending on the device, but that would require a larger rewrite which can also affect the desktop screens )

Issue #486, opened 2023-05-25
Mobile Responsiveness: Toggle Dark/Light mode button doesn't work


Pranav2612000:
The actual hotspot for the button to work is beside the actual location of the button.

joelcmk:
I can help with this issue!

PR #490, opened 2023-05-26
Fix: toggle dark/light mode on mobile

betty-alagwu:
This PR is the first pass for issue #486

Fixes the toggle dark/light mode on mobile.

Here's how it looks 

https://github.com/Mirrorful/mirrorful/assets/94234459/174d8a7a-1fcf-49b2-9b0b-b18114ecc09b


betty-alagwu:
> 

@alexdanilowicz thanks. The `onClick` event handler is on 4 elements: The light icon, the switch, the dark icon, and the container that wraps all three of them. Clicking caused the handler to be called twice. 

The fix here was removing the handler from the container, ensures that only the specific elements within the `Flex` container trigger their respective onClick event handlers.

Thanks for checking and reviewing.

Issue #472, opened 2023-05-24
Enter on color picker should save the value even when still in focus 


alexdanilowicz:
**Steps to repro:**
1. enter a color
2. while still focused on the input, hit enter
3. your color will not be saved
Expected behavior: enter saves your value. 

https://github.com/Mirrorful/mirrorful/assets/29822597/72fa52d1-2213-4b8b-b0c9-5052d63fda19

Pranav2612000:
I'll like to work on this.

noobCode-69:
Since it's been two days , I should help!!

noobCode-69:
If you look at this code ColorPicker.tsx
```js
onChange={(e) => {
    setHexInputValue(e.target.value)
}}
onBlur={() => {
    /* Since the user may not have entered a valid hexcolor string, we parse it to convert it to a valid hexcolor string */
    const parsedHexString = parseHexString(
        hexInputValue,
        tinycolor(colorPickerColor).toHexString()
     )
     setHexInputValue(parsedHexString)
     onChange(parsedHexString)
}}
```
**The behaviour we are expecting** 
- User write the hex value (onChange get triggered updating the local state)
- he will click to the Submit button (onBlur get triggered converting the hexInputValue to valid one and then updating the primaryColor from the index.ts)

**Error**
- When the user directly press "Enter" , the onBlur never get a chance to get triggered

noobCode-69:
Hey @alexdanilowicz please let me know if this works!!

alexdanilowicz:
Sorry I missed this @Pranav2612000 . @noobCode-69 looks good! will merge today.

apologies as this slipped by my notifications

PR #491, opened 2023-05-27
fixed: Enter on color picker should save the value even when still in focus

noobCode-69:
Please review This issue is related to pull request #472

### DIFF
--- a/packages/core/src/components/ColorPalette/ColorPicker.tsx
+++ b/packages/core/src/components/ColorPalette/ColorPicker.tsx
@@ -22,15 +22,18 @@ type TFormat = 'HEX' | 'RGBA'
 interface Props {
   colorPickerColor: AnyColor
   onChange: (color: AnyColor) => void
+  onUpdatePage: (page: number) => void
 }
 
 function DropdownInput({
+  onUpdatePage,
   onChange,
   colorPickerColor,
   format,
   onSetFormat,
   shadow = false,
 }: {
+  onUpdatePage: (page: number) => void
   colorPickerColor: AnyColor
   onChange: (color: AnyColor) => void
   format: TFormat
@@ -108,7 +111,9 @@ function DropdownInput({
           }}
           onKeyDown={(e) => {
             if (e.key === 'Enter') {
+              e.preventDefault()
               updatePrimaryColorToValidColor()
+              onUpdatePage(2)
             }
           }}
         />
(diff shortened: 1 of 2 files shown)


alexdanilowicz:
Hey @noobCode-69, this doesn't seem to work. I press enter and it removes the focus instead of going to the next screen.

noobCode-69:
Ok ,
The problem was the on input when I add a event listener of key down the default behavior of submitting the form on press Enter goes away , had to navigate manually

noobCode-69:
Shit I broke the website!

noobCode-69:
Hey @alexdanilowicz the behaviour that pressing enter on input will submit the form is very difficult to achieve (atleast for me) , since the ColorPicker.tsx is  reused several place

noobCode-69:
And even if you achieve it. The behavior is not very smooth 
- Suppose the ColorPicker.tsx is used in a form where there are several input , then pressing enter will directly submit the form even though you want to change other inputs also.
- Suppose the user paste a color and press enter , the form directly get submitted , even though the user want to see a preview of the color.

Pranav2612000:
Makes sense @noobCode-69 
I think clearly understanding how the enter inside color picker should work is important.

@alexdanilowicz Can we assume that pressing 'Enter' inside any input field in the color picker should never submit a form the color picker is part of?

packages/core/src/components/ColorPalette/ColorPicker.tsx
```diff
@@ -57,6 +57,15 @@ function DropdownInput({
     setHexInputValue(tinycolor(colorPickerColor).toHexString())
   }, [colorPickerColor])
 
+  const updatePrimaryColorToValidColor = () => {
```
Pranav2612000, line 60: @noobCode-69 Do you think also calling this function while unmounting this component will help to solve the problem?

noobCode-69: Yeah , It could solve it but there are some problems
- The ColorPicker.tsx is having it's local state to store the input
- The index.tsx of the onboarding also have a state that store the input
- The ColorPicker.tsx is updating the state of index.tsx with it's local state when the input get blurred
The solution that you suggest won't work because of the third point.

alexdanilowicz:
@Pranav2612000 not sure what you mean. I think expected behavior is you press "enter" and then goes to the next step.

alexdanilowicz:
Going to close this out due to stale inactivity.

Issue #483, opened 2023-05-25
Mobile Responsiveness: Sidebar to be closed be default on mobile when page loads


Pranav2612000:
Right now, when I visit the mirrorful home page, the sidebar is open by default. In line with how the hamburger icons works across most sites, the sidebar should be closed by default and open up when a user clicks on the hamburger icon

Pranav2612000:
Thoughts? @alexdanilowicz

alexdanilowicz:
I think it should be closed by default _on mobile_ .

This is what you mean, yes? @Pranav2612000

betty-alagwu:
@Pranav2612000 @alexdanilowicz i'd love to work on this

Pranav2612000:
Sure @betty-alagwu ! Go ahead.

lazyfuhrer:
Is it still open? I'd like to give a shot @Pranav2612000

PR #494, opened 2023-05-29
fix: Make sidebar close by default

betty-alagwu:
This PR is the first pass for make the sidebar close by default once the page loads 
#483

Here's how it works

https://github.com/Mirrorful/mirrorful/assets/94234459/2f7736a1-5b23-41f1-ba89-73b99a9f99b2


alexdanilowicz:
Thanks Betty! To clarify, it should close by default _on mobile_ only. On desktop it should still be open!

betty-alagwu:
Oh right will get that fixed

alexdanilowicz:
Hi @betty-alagwu, looks like there are some linting issues. Are you able to fix them?

betty-alagwu:
> Hi @betty-alagwu, looks like there are some linting issues. Are you able to fix them?

Yeah i was able to fix them, the checks are passing now.

@alexdanilowicz

packages/core/src/components/Layout.tsx
```diff
@@ -104,7 +105,9 @@ export default function Layout({
     })
   }
 
-  const [isMobileView] = useMediaQuery('(max-width: 480px)')
+  useEffect(() => {
```
Pranav2612000, line 108: Hey @betty-alagwu ! What does this useEffect do? If it's just for handling dimensions change on a device, I think we don't need it. Not a lot of users changing their device's dimensions, except for us testers.

Would just setting the initial value correctly on Line 37 work?

Something like this 

```
const [isSidebarCollapsed, setIsSidebarCollapsed] = useState(false || isMobileView)
```

I wasn't also able to test this out completely. But I think with this useEffect, it will also prevent the sidebar from opening up again on desktop devices once it has been closed - because it'll be triggered every time `setIsSidebarCollapsed` variable is changed, and set it to false.

betty-alagwu: Hello @Pranav2612000 ! Thanks for flagging this. 

The `useEffect` hook helps keep the sidebar's collapsed state in sync with the device's mobile view status.

Setting the initial value directly will not produce the desired behaviour and this is because the initial value of `isMobileView` might not be available immediately since the `useMediaQuery` hook takes some time to determine if the device is in mobile view or not.

Also while the `useEffect` is updating `setIsSidebarCollapsed`, it does not prevent the sidebar from opening up on desktop, such as when the `isMobileView` changes, the sidebar will reopen on desktop devices.

Here's a short video of how it works now:


https://github.com/Mirrorful/mirrorful/assets/94234459/f27f4627-bc1b-4387-8685-6040fcd26d46

Issue #413, opened 2023-05-11
Complementary palette's opacity should match primary's opacity


alexdanilowicz:
For example, if the primary is rgb(A,B,C, 0.25)... then the palette should be generated as  rgb(X,Y,Z, 0.25).

noobCode-69:
Can you elaborate the issue, like how to reproduce this issue!

alexdanilowicz:
Yes, steps to repro:

1. start with onboarding
2. make primary rgba, and change opacity to less than 100%
3. go through onboarding
4. expected behavior: the rest of pallete (the other colors) also have 0.25 opacity, not 100. 

https://github.com/Mirrorful/mirrorful/assets/29822597/c9d40819-327f-463e-92dc-950904c1fd43

noobCode-69:
I could help with that!

noobCode-69:
Little doubt here.
* How to start onboarding process , it happened when i first started the mirrorful , but not every time.
And yeah please assign it to me! I think i can do it.

PR #501, opened 2023-06-16
🐛 fix: Complementary palette's opacity should match primary's opacity

akindoju:
Hi @alexdanilowicz, please review my PR for the issue on #413


Pranav2612000 (approved):

alexdanilowicz:
Thanks so much for this! I fixed this in a separate repo we are working on, which we will make open-core soon. So this change is now live! Really appreciate it @akindoju. 

Resolves #413

